PR #10, opened 2023-09-19
Linting & mypy updates

Kilo59:
Update to latest version of `ruff` + `mypy` and apply latest oss static analysis standards.

Make compatible with pydantic v2

Update `great_expectations` minimum pin to `0.17.17`


tasks.py
```diff
@@ -35,7 +35,7 @@ def type_check(ctx: Context, install_types: bool = False, check: bool = False):
     if install_types:
         cmds.append("--install-types")
     if check:
-        cmds.append("--pretty")
+        cmds.extend(["--pretty", "--warn-unused-ignores"])
```
Kilo59, line 38: If type-ignores are not being used we should remove them

Kilo59:
- See #12 for the CI fix

cdkini (approved):

PR #14, opened 2023-09-20
setup build steps and publish to pypi

allensallinger:
setup build steps and publish to pypi

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,10 +39,6 @@ jobs:
 
       - run: poetry run invoke ${{ matrix.task }} --check
 
-      - name: Build distributions
-        run: poetry build
-
-
   tests:
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -39,6 +39,10 @@ jobs:
 
       - run: poetry run invoke ${{ matrix.task }} --check
 
+      - name: Build distributions
+        run: poetry build
```
Kilo59, line 43: Can you move this to it's own job.
We want these static analysis steps to be fast and only fail if the particular check fails.

Don't want to have the `formatting`/`fmt` step failed because of a build issue.

pyproject.toml
```diff
@@ -26,6 +26,11 @@ tomli = "^2.0.1"
 [tool.poetry.scripts]
 gx-agent = 'great_expectations_cloud.agent:run_agent'
 
+[tool.poetry-dynamic-versioning]
+enable = true
+vcs = "git"
+style = "semver"
```
Kilo59, line 32: I thought we were going to leave this out initially and do things manually.

Kilo59 (changes requested):
Couple of questions

Kilo59 (approved):

PR #18, opened 2023-09-25
Draft: add docker lint

rreinoldsc:
Added docker lint, and addressed the following:

```
-:1 DL3029 warning: Do not use --platform flag with FROM
-:9 DL3013 warning: Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>` or `pip install --requirement <requirements file>`
-:9 DL3042 warning: Avoid use of cache directory with pip. Use `pip install --no-cache-dir <package>`
```


Kilo59 (approved):

great_expectations_cloud/agent/Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM --platform=linux/amd64 python:3.10.12-slim as base
```
rreinoldsc, line 1: not multistage, so no need for alias

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,11 @@
...
+on: [pull_request]
+jobs:
+  # Can look up error with https://github.com/hadolint/hadolint/wiki/DL3029
+  docker-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+      - uses: actions/checkout@v3
+
+      - run: docker run -v --rm -i hadolint/hadolint hadolint --failure-threshold warning - < great_expectations_cloud/agent/Dockerfile
```
Kilo59, line 11: If you put this into an `invoke` task you can make it part of our existing `lint` matrix.

https://github.com/great-expectations/cloud/blob/fb172df6ebb6b1bca86f1a465aa50959da0ed923/.github/workflows/ci.yaml#L12

rreinoldsc: yep all set

PR #19, opened 2023-09-25
SQL Dependency Group

Kilo59:
Add a `sql` poetry dependency group that includes all the required dependencies for `postgres`, `redshift` and `databricks`


pyproject.toml
```diff
@@ -13,6 +13,13 @@ pydantic = "<3"
 pika = "^1.3.1"
 snowflake-sqlalchemy = ">=1.5.0"
 
+[tool.poetry.group.sql.dependencies]
+great-expectations = { version = "^0.17.19", extras = [
+    "databricks",
+    # includes redshift, postgres, etc.
+    "sqlalchemy1"
```
Kilo59, line 20: This was failing before because the `sqlalchemy` gx extra requires `SQLALchemy` 2, which is not compatible with all of our datasources yet.

allensallinger (approved):

PR #20, opened 2023-09-26
add arm64 support

rreinoldsc:
Unable to run on arm64 mac:

```
$ docker image pull greatexpectations/agent:0.0.1
0.0.1: Pulling from greatexpectations/agent
no matching manifest for linux/arm64/v8 in the manifest list entries
```

Adds explicit platform targets for amd64 and arm64


rreinoldsc:
alrdy fixed elsewhere

PR #21, opened 2023-09-26
revert platform tag

rreinoldsc:
not trivial fix to add support for arm64

I removed platform tag in previous [PR]( https://github.com/great-expectations/cloud/pull/18/files#diff-06925c091c7ca7abfba230be53583ab49f0a56b5869b8e708681a8990d3bd3e6L1), which broke docker builds on apple silicon.

This reverts the change, but also reverts us back to emulating the agent on apple silicon.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
   static-analysis:
     strategy:
       matrix:
-        task: ["lint", "fmt", "type-check", "python-build", "docker --check"]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
   static-analysis:
     strategy:
       matrix:
-        task: ["lint", "fmt", "type-check", "python-build", "docker --check"]
```
rreinoldsc, line 12: this was redundant, --check gets called below

PR #24, opened 2023-09-27
[MAINTENANCE] Make `great_expectations` dependency unbounded

cdkini:
Due to dependency resolver issues, we should accept ANY version of GX (although >=17 is actually what is required). We can resolve the versioning issues in OSS later on

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.4dev"
+version = "0.0.3dev1"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,14 +1,14 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.3dev"
+version = "0.0.4dev"
```
Kilo59, line 3:
```suggestion
version = "0.0.3dev1"
```

Kilo59 (approved):

PR #31, opened 2023-10-03
publish to pypi gx-cloud

allensallinger:
publish to pypi gx-cloud

### DIFF
--- a/.github/workflows/publish-to-pypi-gx-cloud.yaml
+++ b/.github/workflows/publish-to-pypi-gx-cloud.yaml
@@ -1,4 +1,4 @@
-name: publish-to-pypi
+name: publish-to-pypi-gx-cloud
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi-gx-cloud.yaml
Kilo59, line 1: How does this action know to publish to `gx-cloud` instead of `great-expectations-cloud`?

allensallinger: My current understanding is that is authenticates via the mix of an org name, a repo name and the workflow that you specify. I chose to name it a different workflow for now in case there were collisions.

Shinnnyshinshin:
Hi @allensallinger and @Kilo59 wanted to check. Is this PR still needed?

wookasz:
this is over a year old, closing it as stale

PR #41, opened 2023-10-20
Add version bump developer `invoke` task

Kilo59:
Add task for bumping the poetry/pypi version

- [Add `tomlkit` dev dependency ](https://tomlkit.readthedocs.io/en/latest/)for reading/writing from `.toml` files
  - `tomli-w` does not  (as of now) preserve formatting or comments, so not a good choice for this task

```console
$ invoke version-bump --help
Usage: inv[oke] [--core-opts] version-bump [--options] [other tasks here ...]

Docstring:
  Bump project version.

Options:
  -p, --pre        Bump the pre-release version (Default)
  -s, --standard   Bump the non pre-release micro version
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,8 +36,7 @@ pytest-icdiff = "*"
 pytest-mock = "*"
 responses = "^0.23.1"
 ruff = "^0.1.0"
-tomli = "^2.0.1"
-tomli-w = "^1.0.0"
+tomlkit = "^0.12.1"
 typing_extensions = ">=4.4.0"
 
 [tool.poetry.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 name = "great_expectations_cloud"
 version = "0.0.3dev5"
 description = "Great Expectations Cloud"
-authors = ["Your Name <you@example.com>"]
+authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 readme = "README.md"
 license = "Apache-2.0"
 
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -90,7 +120,79 @@ def build(ctx: Context) -> None:
         "-t",
         "greatexpectations/agent",
         "-f",
-        DOCKERFILE_PATH,
+        str(DOCKERFILE_PATH),
```
Kilo59, line 123:
```suggestion
        DOCKERFILE_PATH,
```

pyproject.toml
```diff
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.3dev5"
```
Kilo59, line 3: Explicitly adding the `.` between the `micro` and `pre-release` components.
https://pypi.org/project/great-expectations-cloud/0.0.3.dev5/

allensallinger (approved):
A few questions though. For major or minor releases will those be done manually? And do we want to have this as part of CI so releases are cut automatically or will this be a manual step as well?

tasks.py
Kilo59, line 1: >A few questions though. For major or minor releases will those be done manually?

Yes, major version releases would be done manually.


> And do we want to have this as part of CI so releases are cut automatically or will this be a manual step as well?

Eventually yes. Although I think this should be lower priority than other things we need like...
1. automatic changelogs
2. creating Github releases
3. Coming up with Docker tagging/release strategy for pre-releases

PR #44, opened 2023-10-23
Release 0.0.3

Kilo59:
Fix pre-release -> standard release logic.

The next release after a pre-release is a non-pre-release version.
```
0.0.1.dev1 -> 0.0.1 -> 0.0.2.dev0 -> 0.0.2dev1 -> 0.0.2
```

Followup to
 - https://github.com/great-expectations/cloud/pull/41

### Note
The actual `0.0.3` release could be cut separately if desired.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,8 @@
 name = "great_expectations_cloud"
 version = "0.0.3"
 description = "Great Expectations Cloud"
+repository = "https://github.com/great-expectations/cloud"
+homepage = "https://greatexpectations.io"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 readme = "README.md"
 license = "Apache-2.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.3.dev8"
+version = "0.0.3"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,8 +2,6 @@
 name = "great_expectations_cloud"
 version = "0.0.3"
 description = "Great Expectations Cloud"
-repository = "https://github.com/great-expectations/cloud"
-homepage = "https://greatexpectations.io"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 repository = "https://github.com/great-expectations/cloud"
 homepage = "https://greatexpectations.io"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.3.dev8"
+version = "0.0.3"
```
Kilo59, line 3: This would make this the first version full/non-pre-release version.
I'm happy to change this back to a pre-release if we don't think we are ready.
But I wanted to get the logging changes into `great_expectations[cloud]` installs.

tasks.py
```diff
@@ -141,13 +141,18 @@ def _get_latest_version() -> Version:
...
     if not pre_release:
-        # standard release
-        new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro + 1}")
+        # standard release - remove the dev component if it exists
+        if version_.dev:
+            new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro}")
+        else:
+            new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro + 1}")
     elif version_.dev:
+        # bump and existing pre-release version
```
Kilo59, line 150:
```suggestion
        # bump an existing pre-release version
```

tasks.py
```diff
@@ -141,13 +141,18 @@ def _get_latest_version() -> Version:
...
         new_version = Version(
-            f"{version_.major}.{version_.minor}.{version_.micro}.dev{version_.dev + 1}"
+            f"{version_.major}.{version_.minor}.{version_.micro}.dev{version_.dev + 0}"
         )
     else:
+        # create the first pre-release version
         new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro}.dev1")
```
Kilo59, line 156:
```suggestion
        new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro}.dev0")
```

tasks.py
```diff
@@ -141,13 +141,18 @@ def _get_latest_version() -> Version:
 
 def _bump_version(version_: Version, pre_release: bool) -> Version:
     if not pre_release:
-        # standard release
-        new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro + 1}")
+        # standard release - remove the dev component if it exists
+        if version_.dev:
+            new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro}")
```
allensallinger, line 146: Is this var still in scope when it is referenced below since it is not first available outside the if/else statements. And did you test this locally first?

Kilo59: It's not in scope in the else blocks but each block is creating initializing a `new_version` object.
`mypy` would catch it if I was forgetting a case and not returning a `Version` object.

Yes I've tested it locally.

The reason I'm changing it isn't because the function didn't work but because it was doing the wrong thing.
For example `0.0.3dev1` comes before `0.0.3`, not after it.

allensallinger (approved):

PR #45, opened 2023-10-24
Add `orjson` dependency to support Descriptive Metrics

Kilo59:
Our Descriptive Metrics feature relies on `orjson`'s ability to serialize numpy types.
It should be a core dependency of the `great_expectations_cloud` package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,8 @@ name = "great_expectations_cloud"
 version = "0.0.3.dev8"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
+repository = "https://github.com/great-expectations/cloud"
+homepage = "https://greatexpectations.io"
 readme = "README.md"
 license = "Apache-2.0"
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,8 +1,10 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.3.dev8"
+version = "0.0.3.dev9"
```
allensallinger, line 3: Do these pre release tags reset each time a minor version goes out?

allensallinger (approved):

pyproject.toml
Kilo59: Yes.
You should play around with it locally.

Imagine we are on `0.0.4`, and we choosing between a new `micro` version or a new pre-release version.

1. New micro version (non-pre-release)
```console
$ invoke version-bump  --standard
local:          0.0.4
pypi latest:    0.0.3

  bumping version ... 
new version:    0.0.5

✅ 0.0.5 will be published to pypi on next merge to main
```
2. New pre-release version
```console
invoke version-bump            
local:          0.0.4
pypi latest:    0.0.3

  bumping version ... 
new version:    0.0.4.dev1

✅ 0.0.4.dev1 will be published to pypi on next merge to main
```

PR #46, opened 2023-10-24
Ruff formatter

Kilo59:
Adopt the `ruff` formatter and drop `black`

https://astral.sh/blog/the-ruff-formatter


wookasz (changes requested):
Let's figure out which tools we'd like to use, and use them uniformly across our repositories to avoid cognitive overhead.

Kilo59:
@wookasz 
I'm proposing that we switch to using `ruff` `format` uniformly.
Assuming no one objects and nothing unexpected pops up in the attempt to adopt it here.
Do you have a specific objection to switching to the `ruff` formatter?

Developers will likely not even notice the difference.

1. If developers have the `pre-commit` hooks installed the formatting is applied automatically when committing code.
2. If developers are using `invoke fmt` the only difference is `invoke` calls `ruff format` instead of `black`
3. If developers don't use either and misformat something the pre-commit CI will push a fix to the branch automatically.

AFAIK the only case where a developer workflow would change is if they are manually running `black` from the CLI after every change.

wookasz (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.7.0 and should not be changed by hand.
```
Kilo59, line 1: Using a new version of `poetry` here.

PR #47, opened 2023-10-26
Keep `pre-commit` and `poetry` linters in sync

Kilo59:
`pre-commit` autoupdate can lead to our pre-commit version getting ahead of our pinned versions of `ruff` and `black`.
Update `ruff` & `black` to use the latest version and add a test to prevent this problem in the future.


pyproject.toml
```diff
@@ -166,3 +166,4 @@ markers =[
     "agentjobs",
     "unit",
 ]
+log_level = "info"
```
Kilo59, line 169: Updating the default log level for `pytest` to `info`.
Makes failures easier to debug.

roblim (approved):

PR #49, opened 2023-10-26
Update LICENSE

wookasz:
This project is not fully open, updating the licnese


Kilo59:
The pypi project metadata comes from `pyproject.toml` so we need to update it there as well

Kilo59 (approved):

PR #50, opened 2023-10-26
Fix and test version bumping logic

Kilo59:
Previous logic when bumping a standard release to a `pre-release` was wrong.
It would result in the version going backward. `0.03` -> `0.0.3.dev0` instead of `0.0.3` -> `0.0.4.dev0`

Added tests to verify this behavior this time.


tasks.py
```diff
@@ -153,7 +153,7 @@ def _bump_version(version_: Version, pre_release: bool) -> Version:
         )
     else:
         # create the first pre-release version
-        new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro}.dev1")
+        new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro + 1}.dev0")
```
allensallinger, line 156: Question: why do we call the third version a `micro`? Generally I have heard semver referenced as `major.minor.match`

Kilo59: This isn't our code.
It's coming from `Version` which handles other types of versioning formats.
https://packaging.pypa.io/en/stable/version.html#packaging.version.Version

allensallinger (approved):

PR #56, opened 2023-11-08
add container smoketest check in CD

rreinoldsc:


rreinoldsc:
nvm I see it is tested prior

PR #60, opened 2023-11-09
enable nightly builds

rreinoldsc:
- Add nightly trigger, so we pick up latest GX without requiring a merge into main
- Document a few steps

### DIFF
--- a/.github/workflows/containerize-agent.yaml
+++ b/.github/workflows/containerize-agent.yaml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - main
+  schedule:
+    # Nightly build, midnight ET
+    # https://crontab.guru/#0_5_*_*_*
+    - cron:  '0 5 * * *'
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/containerize-agent.yaml
```diff
@@ -55,14 +59,17 @@ jobs:
...
-      - name: Build Test Image
+      # https://docs.docker.com/build/ci/github-actions/test-before-push/
+      - name: Build, No Push
         uses: docker/build-push-action@v4
         with:
           context: .
           platforms: linux/amd64
+          push: false
           load: true
           tags: greatexpectations/agent:latest
```
allensallinger, line 70: Should we tag this as a nightly build?

rreinoldsc: At this point, no. It has no functional differences from `latest`, and we're already retagging a semver

allensallinger (approved):

PR #64, opened 2023-11-15
fix import error snowflake

rreinoldsc:
Agent starts up, but when snowflake jobs starts, fails with:
```
ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
...
snowflake
```
Our poetry.lock showed urllib>1, but GX is indirectly on 2.0.4. So I set a reasonable >2 in our reqs.

Synthetic tests showing failures: https://app.datadoghq.com/logs?query=service%3Agx-agent%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1699486806885&to_ts=1700091606885&live=true


wookasz (approved):

rreinoldsc:
repro proved this didnt fix it, need to dig further

rreinoldsc:
The issue was resolved another way. deleting

PR #65, opened 2023-11-16
Roll back poetry lock issue

rreinoldsc:
poetry.lock file shows a regression in the snowflake connector. Unclear what change in the pyproject.toml 

image

Repro before and after, and this addresses the issue:

Agent starts up, but when snowflake jobs starts, fails with:
```
ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
```


rreinoldsc:
Concerned that this also rolls back GX 0.18.1 to 0.17.9

Kilo59 (changes requested):
If the issue is with a particular package we should adjust that package version, we don't need to roll back these unrelated commits.

To debug what's causing the package version differences you can use `poetry show --tree` to render a dependency graph.
Or `poetry show <PACKAGE_NAME>` to show information about just that dependency.
```shell
poetry show snowflake-connector-python       
 name         : snowflake-connector-python     
 version      : 2.0.2                          
 description  : Snowflake Connector for Python 

dependencies
 - asn1crypto >0.24.0,<2.0.0
 - azure-common *
 - azure-storage-blob *
 - boto3 >=1.4.4,<1.10.0
 - botocore >=1.5.0,<1.13.0
 - certifi *
 - cffi >=1.9
 - cryptography >=1.8.2
 - future *
 - idna *
 - ijson *
 - oscrypto <2.0.0
 - pycryptodomex >=3.2,<3.5.0 || >3.5.0
 - pyjwt *
 - pyOpenSSL >=16.2.0
 - pytz *
 - six *

required by
 - snowflake-sqlalchemy <4.0.0
```

PR #66, opened 2023-11-16
add `snowflake-connector-python` constraint and re-lock deps

Kilo59:


poetry.lock
```diff
@@ -3706,45 +3415,57 @@ files = [
 
 [[package]]
 name = "snowflake-connector-python"
-version = "2.0.2"
+version = "3.5.0"
```
rreinoldsc, line 3418: lg

rreinoldsc (approved):

PR #71, opened 2023-11-29
Ssl handshake debugging

nicgrayson:


Kilo59 (approved):
You'll need to bump the version in order for this to publish to pypi
```console
invoke version-bump
```

Shinnnyshinshin (commented):
Thank you so much for this @nicgrayson

PR #75, opened 2023-12-05
unpin GX from 0.18.2; pin to latest

JennyTee:
Reverting GX pin to 0.18.2 (added in #73) now that issue preventing agent from using GX 0.18.3 has been resolved.


.pre-commit-config.yaml
```diff
@@ -11,7 +11,7 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, develop, --branch, main]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.1.6"
+    rev: "v0.1.7"
```
JennyTee, line 14: Running `poetry update` also updated the version of ruff we use. Without this change the `test_pre_commit_versions_are_in_sync` test was failing

tasks.py
```diff
@@ -24,7 +24,7 @@
 
 @functools.lru_cache(maxsize=8)
 def _get_pyproject_tool_dict(
-    tool_key: Literal["poetry", "black", "ruff", "mypy"] | None = None
```
JennyTee, line 27: Without this change the action to run `poetry run invoke fmt --check` was failing

PR #77, opened 2023-12-07
Add docker badge

Kilo59:


JennyTee (approved):
👍

JennyTee:
> Alternatively, we could use one of these other Docker badges. But we might want to wait until we've gotten our install size down?? https://shields.io/badges/docker-image-size-with-architecture-latest-by-date-latest-semver https://shields.io/badges/docker-image-size-tag

I think this is good for now but yeah maybe we consider the other ones at a later time.

PR #80, opened 2023-12-13
F/zel-190/zel-220/write logs to local file

JennyTee:
Write detailed logs to log files--if users encounter unexpected errors, they can send the log file(s) to GX for help identifying/resolving the issue.

If a custom configuration JSON file is provided to the agent, it will use that instead. No log files will be generated (unless the config file specifies they should be).

If a custom config file isn't provided, log files will be created. Note: the Agent will still accept a `--log-level` argument and will print to the terminal as needed based off of that argument (in addition to writing to log files).

With this implementation, a new timestamped log file is generated daily at midnight. Up to 30 daily logs are kept at once on a rotating basis.

I did not remove/update any of the `print` statements in the code--I think we may want to separate user prompts and messages with the actual logging info.

image
Note: `logfile` is the initial file created upon agent startup--it's not possible to timestamp the initial file.

image


README.md
JennyTee, line 1: I started off just updating "agent " to "the GX Agent"; ended up updating some other language/style/grammar for consistency.

great_expectations_cloud/agent/agent.py
```diff
@@ -64,13 +64,13 @@ class GXAgent:
...
-    To start the agent, install Python and great_expectations and run `gx-agent`.
-    The agent loads a DataContext configuration from GX Cloud, and listens for
+    To start the GX Agent, install Python and great_expectations and run `gx-agent`.
+    The GX Agent loads a DataContext configuration from GX Cloud, and listens for
     user events triggered from the UI.
     """
 
     def __init__(self: Self):
-        print("Initializing GX-Agent")
+        print("Initializing the GX Agent")
```
JennyTee, line 73: Since we have the debug logs, I don't think it's necessary to also log this to the log files--but I can update that if others disagree. I thought we may want to separate user prompts/info from the debug logs.

great_expectations_cloud/logging_cfg.py
```diff
@@ -54,5 +54,27 @@ def configure_logger(log_level: LogLevel, log_cfg_file: pathlib.Path | None) ->
...
     else:
-        logging.basicConfig(level=log_level.numeric_level)
-        LOGGER.info(f"Configured logging at level {log_level}")
+        logDirectory = pathlib.Path("logs")
+        if not logDirectory.exists():
+            pathlib.Path(logDirectory).mkdir()
+
+        logger = logging.getLogger()
+        formatter = logging.Formatter(
+            "%(asctime)s | %(name)s | %(lineno)d | %(levelname)s: %(message)s"
```
JennyTee, line 63: See screenshot in the description of this PR to see how this turns out. I thought it was nice to have the line number in there, too.

great_expectations_cloud/logging_cfg.py
```diff
@@ -54,5 +54,27 @@ def configure_logger(log_level: LogLevel, log_cfg_file: pathlib.Path | None) ->
...
+        )
+        logger.setLevel(logging.DEBUG)
+
+        # The StreamHandler writes logs to stderr based on the provided log level
```
JennyTee, line 67: In order to write to both stderr and the file, I added 2 logging handlers--one that writes to stderr and shows up on the terminal and one that writes to the files. 

Maybe eventually we can simplify to having just the files, but I think it's helpful to have a user-configurable logging level for output to stderr.

great_expectations_cloud/logging_cfg.py
anthonyburdi: Agree on including the line number :) One question: Is it customary to have the number just by itself (e.g. for easier parsing)? Or have a `line:` or some other prefix before the line number?

README.md
```diff
@@ -10,11 +10,11 @@
 
 1. [Install `poetry`](https://python-poetry.org/docs/#installation)
    - [`pipx install poetry`](https://python-poetry.org/docs/#installing-with-pipx)
-2. Setup virtual environment and install dependencies.
+2. Set up virtual environment and install dependencies.
```
anthonyburdi, line 13: What do you think about moving these README changes into a separate PR? Just in case we have to roll back the logging changes, it would be nice to not also have to roll back the README changes. Since this is an internal facing repo (the README isn't seen by users) I won't block the PR on this point.

anthonyburdi (commented):
Looks good to me! Just a few questions.

great_expectations_cloud/logging_cfg.py
JennyTee: I like the idea of using `line:`; that makes it easier to understand--will update, thank you!

README.md
JennyTee: deal, fair point--will add to separate PR, ty

JennyTee: spinoff PR is here: https://github.com/great-expectations/cloud/pull/81

anthonyburdi: Thank you for separating! ✅  on spinoff.

.gitignore
```diff
@@ -152,6 +152,9 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+# Agent-generated log files
+logs/
```
rreinoldsc, line 156: Will you ensure this is also ignored in the .dockerignore so we dont accidentally build any sensitive logs into the image

great_expectations_cloud/logging_cfg.py
```diff
@@ -54,5 +55,27 @@ def configure_logger(log_level: LogLevel, log_cfg_file: pathlib.Path | None) ->
...
-        LOGGER.info(f"Configured logging at level {log_level}")
+        logDirectory = pathlib.Path("logs")
+        if not logDirectory.exists():
+            pathlib.Path(logDirectory).mkdir()
+
+        logger = logging.getLogger()
+        formatter = logging.Formatter(
+            "%(asctime)s | %(name)s | line: %(lineno)d | %(levelname)s: %(message)s"
+        )
+        logger.setLevel(logging.DEBUG)
```
rreinoldsc, line 66: we have a `--log-level` flag that should be honored

great_expectations_cloud/logging_cfg.py
```diff
@@ -54,5 +55,27 @@ def configure_logger(log_level: LogLevel, log_cfg_file: pathlib.Path | None) ->
...
+        stream_handler.setLevel(log_level.numeric_level)
+        stream_handler.setFormatter(formatter)
+        logger.addHandler(stream_handler)
+
+        # The FileHandler writes all logs to a local file
+        file_handler = logging.handlers.TimedRotatingFileHandler(
+            filename=logDirectory / "logfile", when="midnight", backupCount=30
+        )  # creates a new file every day; keeps 30 days of logs at most
+        file_handler.setFormatter(formatter)
+        file_handler.setLevel(logging.DEBUG)
```
rreinoldsc, line 79: same as above

great_expectations_cloud/logging_cfg.py
JennyTee: So apparently the root logger needs to be set up with a log level that is equal to or lower than the `FileHandler` and `StreamHandler`.  So I made the root logger use the lowest option.

Then the `StreamHandler` respects the user-provided `--log-level` (or defaults to `logging.WARN` if the user didn't provide the a `--log-level`.

I was assuming that we want the `FileHandler` to show the most verbose logging level possible, which is why I set it to `logging.DEBUG`.  If it defaults to just `logging.WARN`, will it give us enough info to debug issues? If so, I can update.

.gitignore
JennyTee: ah thanks - we didn't have a .dockerignore file yet so I added one

great_expectations_cloud/logging_cfg.py
JennyTee: tks - per comment above (https://github.com/great-expectations/cloud/pull/80#discussion_r1425964327), we will keep as logging.DEBUG to make sure that the log file has all of the debug logging. But the StreamHandler that prints to stderr in the terminal will respect the `--log-level` param

great_expectations_cloud/logging_cfg.py
JennyTee: Actually after more testing, the root logger `setLevel` is actually needed. I thought it was working without, but after more testing I realized that without setting the level of the root logger, it blocks DEBUG and INFO logs to the file. So I added it back and added a comment explaining why.

rreinoldsc:
Add'l documentation only when triaging would include something like:

```
docker run ... -v $(pwd)/logs:/app/logs ...
```

great_expectations_cloud/agent/cli.py
```diff
@@ -5,14 +5,15 @@
...
-from great_expectations_cloud.logging_cfg import LogLevel, configure_logger
+from great_expectations_cloud.logging.logging_cfg import LogLevel, configure_logger
 
 LOGGER = logging.getLogger(__name__)
 
 
 @dc.dataclass(frozen=True)
 class Arguments:
     log_level: LogLevel
+    skip_log_file: bool
```
JennyTee, line 16: Added this param after convo with @nicgrayson

great_expectations_cloud/logging/README.md
JennyTee, line 1: Figured a readme might be helpful, too

JennyTee:
> Suggestion: Add'l documentation only when triaging would include a volume mount something like:
> 
> ```
> docker run ... -v $(pwd)/logs:/app/logs ...
> ```
> 
> Antisuggestion:
> 
> * Using a `docker volume`, bc statefulness

@rreinoldsc Thanks for chatting with me about this! After discussion with @nicgrayson we were thinking that the best solution would be to use the docker logs (which persist after a container shuts down, just won't include debug logs) and keep the functionality that saves rotating debug logs inside the container while it is up. This avoids the complexity of a bind mount or volume. In local tests, even when I encountered warnings/errors, the container was still up, so I could go in and grab the logfiles if needed.

I added a README.md file that explains the different logging options, as well as how to access the logs, if needed.

great_expectations_cloud/logging/README.md
```diff
@@ -0,0 +1,32 @@
...
+Users can provide an optional `--skip-log-file` argument to skip saving log output to files if desired.
+
+A new log file is generated each day with a timestamped filename. The most recent 30 files are kept on a rotating basis.
+
+These files are saved to a `logs` directory in the Docker container and are not accessible after a container has been exited.
+
+Users can access the `logs` directory inside of a running container with the following command(s):
+
+* To copy the logs directory to your local machine: `docker cp {dockerImageId}:/app/logs ~/Desktop/logs`
+* To access the logs directory in the container: `docker exec -it {dockerImageName} /bin/bash`
```
JennyTee, line 26: @nicgrayson , I know you had mentioned giving users a command to copy all files into one file to send to GX, but since the filenames are timestamped, I think it's better to just use the `docker cp` command. But I left in the `exec -it` command in case it's helpful

Shinnnyshinshin (approved):
Thank you for the adding the README and for working outloud for the ongoing discussions. Would like to defer final approval to those who have been part of the ongoing discussions

anthonyburdi (approved):
LGTM!

PR #83, opened 2023-12-16
[MAINTENANCE] Verify version bump

anthonyburdi:
Verify that the version is bumped if the agent code changes. 

Now that the version on [docker hub](https://hub.docker.com/r/greatexpectations/agent) matches the pypi version (specified in pyproject.toml), we want to make sure that this version is bumped each time the agent code is changed. This test fails if the version has not changed from the previously released version.

Test with no files changed does not check for version bump:
https://github.com/great-expectations/cloud/actions/runs/7228373527
image

Test showing failure when the version has not been updated:
https://github.com/great-expectations/cloud/actions/runs/7228211346/job/19697241788
image

See also source ticket: https://greatexpectations.atlassian.net/browse/ZEL-235

Future improvements not included in this PR can include parsing the versions to check if the new version is greater than the old one. For now we can manually check this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
 
   check-if-agent-changed:
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
+    if: ${{ github.event_name != 'schedule' }}
     outputs:
       agent_changed: ${{ steps.check_agent.outputs.agent_changed }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,6 +6,68 @@ on:
     - cron: "0 */2 * * *"
```
rreinoldsc, line 6: no need for cron

rreinoldsc (approved):

.github/workflows/ci.yaml
anthonyburdi: Good catch. Do you mean that I should remove in this PR? I think so since it may fail due to the versions not being different.

rreinoldsc: yeah rm in this PR

anthonyburdi: Left in bc it may be useful for OSS releases that are not in sync with agent released. E.g. if OSS releases a change that breaks the agent, this scheduled test will fail and will run even if we aren't making changes to the agent. I did make a change to only run the version bump jobs if we are in a PR though (and not on schedule) so that only the CI tests run on the schedule.

rreinoldsc: ohh this is for all of CI. Sorry I thought you had added it. yeah makes sense to leave

anthonyburdi: All good - great catch. The version bump check would have failed on scheduled runs.

PR #54, opened 2023-11-03
[MAINTENANCE] Make unfiltered warnings into errors

anthonyburdi:
To make sure unhandled warnings don't show when users are running our agent, this change will cause CI to fail in a PR if a warning is raised during any test.

PR #86, opened 2023-12-20
Ensure Poetry lockfile was generated with a recent version of Poetry

Kilo59:
Adds a test to ensure the version of `Poetry` used to generate the `poetry.lockfile` is reasonably up to date.
We've had a couple of recent PR's where the version of poetry used for the lockfile shifted between `1.6` -> `1.7` and back again.

When poetry makes a new release, we will see a warning in the test output, but it won't fail the build unless it is older than the hard-coded value we test against.
image


pyproject.toml
```diff
@@ -176,7 +176,7 @@ markers =[
 ]
 log_level = "info"
 filterwarnings = [
-    #"error",  # Turn all warnings not explicitly filtered below into errors
+    "error",  # Turn all warnings not explicitly filtered below into errors
```
Kilo59, line 179: This should have been uncommented in this PR
 - https://github.com/great-expectations/cloud/pull/54

pyproject.toml
```diff
@@ -176,7 +176,7 @@ markers =[
 ]
 log_level = "info"
 filterwarnings = [
-    #"error",  # Turn all warnings not explicitly filtered below into errors
+    # "error",  # Turn all warnings not explicitly filtered below into errors
```
Kilo59, line 179:
```suggestion
    #"error",  # Turn all warnings not explicitly filtered below into errors
```

pyproject.toml
rreinoldsc: cc @Shinnnyshinshin

Kilo59: Resolved 
- https://github.com/great-expectations/cloud/pull/88

tests/test_project.py
```diff
@@ -96,5 +98,51 @@ def test_bump_version(version_initial: Version, expected_version: Version, pre_r
...
+            category=PoetryVersionOutdated,
+            stacklevel=1,
+        )
+    return latest_version
+
+
+def test_lockfile_poetry_version(lock_file_poetry_version: Version, latest_poetry_version: Version):
+    """
+    This test ensures that the poetry.lock file was generated using a recent version of poetry.
+    This is important because the lockfile format or dependency resolving strategery could change between versions.
```
JennyTee, line 138:
```suggestion
    This is important because the lockfile format or dependency resolving strategy could change between versions.
```

JennyTee (approved):
This is cool--thanks!

tests/test_project.py
Kilo59: 🤣 ... thank you!

PR #87, opened 2023-12-20
Bump Docker GX OSS version to `0.18.6`

Kilo59:
Critically, this version includes changes to the default value for the `SQLDatasource.create_temp_table` field from `True` to `False` and a related fix for `QueryAsset` so that it works without `temp_tables`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -752,13 +752,13 @@ files = [
 
 [[package]]
 name = "great-expectations"
-version = "0.18.5"
+version = "0.18.6"
 description = "Always know what to expect from your data."
 optional = false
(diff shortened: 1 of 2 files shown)


poetry.lock
```diff
@@ -752,13 +752,13 @@ files = [
 
 [[package]]
 name = "great-expectations"
-version = "0.18.5"
+version = "0.18.6"
```
Kilo59, line 755: GX version change

rreinoldsc: perfect

rreinoldsc (approved):

allensallinger (approved):

PR #88, opened 2023-12-20
`pyproject.toml` update - Warnings are Errors

Shinnnyshinshin:
`filterwarnings` now correctly filters warnings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.14"
+version = "0.0.15.dev0"
 description = "Great Expectations Cloud"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.14"
+version = "0.0.15.dev0"
```
Kilo59, line 3: Why do we need to bump the version here. The code hasn't changed?

Kilo59 (changes requested):

pyproject.toml
Shinnnyshinshin: Ah you're right. Change has been reverted

Kilo59 (approved):

PR #95, opened 2024-01-16
Display GX Cloud Version to user instead of GX warning

Shinnnyshinshin:
* Suppress warning about `great_expectations` version. 
* Add message about current `great_expectations_cloud` version. 

### Note
* This PR doesn't include changes that compare the current version to the `latest` version in Docker, as this will be handled by ZEL-408

### Before
Screenshot 2024-01-16 at 3 30 51 PM
### After
Screenshot 2024-01-16 at 3 29 45 PM


great_expectations_cloud/agent/agent.py
```diff
@@ -111,6 +120,11 @@ def _listen(self) -> None:
             if subscriber is not None:
                 subscriber.close()
 
+    def _get_current_gx_agent_version(self) -> str:
+        version: str = metadata_version(self._PYPI_GX_AGENT_PACKAGE_NAME)
```
anthonyburdi, line 124: Nice! I was going to suggest this (saw the `pkg_resources` version last night).

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.17"
+version = "0.0.18.dev0"
```
anthonyburdi, line 3: Should this be a normal version not a dev version?
```suggestion
version = "0.0.18"
```

anthonyburdi (changes requested):
Looks good, just need an answer on the version before approving.

pyproject.toml
Shinnnyshinshin: Thanks for catching this. I just went with the `poetry` output... I think `0.0.18` is correct.

anthonyburdi: Thanks for changing. I think there is an option to the invoke command for a standard version bump. Might be `--standard` https://github.com/great-expectations/cloud?tab=readme-ov-file#release-to-pypi-and-docker

anthonyburdi (approved):
Looks great, thank you!

great_expectations_cloud/agent/agent.py
JennyTee: Yeah, this is awesome; didn't realize it was that easy!

great_expectations_cloud/agent/agent.py
```diff
@@ -69,11 +70,19 @@ class GXAgent:
     user events triggered from the UI.
     """
 
+    _PYPI_GX_AGENT_PACKAGE_NAME = "great_expectations_cloud"
+
     def __init__(self: Self):
-        print("Initializing the GX Agent")
+        agent_version: str = self._get_current_gx_agent_version()
+        print(f"Initializing the GX Agent version: {agent_version}.")
```
JennyTee, line 77: Oops, didn't realize I added a bullet point for this to the AC; my bad. I originally planned to handle this in https://greatexpectations.atlassian.net/browse/ZEL-301. In any case we are ahead of the game; thanks for doing this. 

I added a note to ZEL-301 about the formatting. I think throwing parentheses around the version info looks nice. No changes requested now; we can handle in that ticket!

Shinnnyshinshin: Thanks for catching this @JennyTee. Yes, looks like we can handle in ZEL-301

great_expectations_cloud/agent/agent.py
```diff
@@ -69,11 +70,19 @@ class GXAgent:
...
         self._config = self._get_config()
         print("Loading a DataContext - this might take a moment.")
-        self._context: CloudDataContext = get_context(cloud_mode=True)
+
+        with warnings.catch_warnings():
+            # suppress warnings about GX version
+            warnings.filterwarnings("ignore", message="You are using great_expectations version")
+            self._context: CloudDataContext = get_context(cloud_mode=True)
```
Kilo59, line 84: Why are we doing this?
Version mismatches between a user's version of `gx` and the `agent`'s version is a common source of issues.

IMO showing the version of `gx-agent` is not a suitable replacement as there's not an easy way for a user to correlate the `gx-agent` version and the `great_expectations` version.

Shinnnyshinshin: Hi @Kilo59 I wanted to follow-up here. I saw that Don was able to share the ticket that we had created for this PR. Does the work that we are intending to do [as a follow-up](https://greatexpectations.atlassian.net/browse/ZEL-275) address your concerns here? If not, happy to talk further on what a good approach would be.

PR #96, opened 2024-01-18
Use postponed annotations

Kilo59:
Force PEP-0563 postponed annotations for all modules.
https://peps.python.org/pep-0563/

This prevents unnecessary imports at runtime, reducing the number of external dependencies and the risk of cyclic imports.

1. Apply `from __future__ import annotations` to every module and enforce this with `ruff`
2. Update ruff `TCH` settings to properly account for pydantic models that cannot use the above postponed annotations.
3. Add `TCH` to the list of ["safe" fixes](https://docs.astral.sh/ruff/linter/#fix-safety). These rules are about moving imports into and out of `if TYPE_CHECKING:` blocks, depending on if they are used at runtime or not. This can be a tedious developer task and is often a source of confusion.
4. Update version of `ruff` to `0.1.13`


cdkini (approved):

pyproject.toml
```diff
@@ -136,23 +136,34 @@ ignore = [
...
     "TID252", # Relative imports from parent modules are banned
     # https://beta.ruff.rs/docs/rules/#flake8-use-pathlib-pth
     "PTH123", # pathlib-open - this would force pathlib usage anytime open or with open was used.
 ]
+# https://docs.astral.sh/ruff/linter/#fix-safety
+extend-safe-fixes = [
+    # pydantic models use annotations at runtime and need to be accounted for
+    # these are safe to apply automatically given settings for tool.ruff.flake8-type-checking
+    "TCH", # https://beta.ruff.rs/docs/rules/#flake8-type-checking-tch
+]
```
Kilo59, line 148: There's a lot of changes in this PR (applied by ruff) but these `ruff` config changes are the important ones.

pyproject.toml
```diff
@@ -168,6 +179,9 @@ max-complexity = 10
     "F401", # unused import
     "F403", # star imports
 ]
+"models.py" = [
+    "UP007", # non-pep604-annotation - pydantic models use annotations at runtime
+]
```
Kilo59, line 184: Not super happy about `UP007` not respecting the `runtime-evlated-base-classes`
I will raise this issue on https://github.com/astral-sh/ruff

PR #97, opened 2024-01-18
Set `User-Agent: gx-agent`& `Agent-Job-Id` on GX Cloud API requests 

Kilo59:
Ensure that a `User-Agent: gx-agent/<AGENT_VERSION>` header is always set when making calls to the GX Cloud REST API.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

Doing this so that we can differentiate between API calls coming from the `gx-agent` vs standard `great_expectations`.

In the context of a job, the `Agent-Job-Id` header is also set on every API request.

### Implementation Notes

The header details are controlled in `great_expectations.core.http` by non-public methods so I had to monkeypatch one of these methods to ensure the header is properly set.

For the V1 release, we should expose a public method of doing this operation, and then this implementation can be changed. I will implement this in a followup so as not to force a new `0.18.x` release.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -298,7 +298,11 @@ class GXAgent:
     def _set_http_session_headers(correlation_id: str | None = None) -> None:
         """
         Set the the session headers for requests to GX Cloud.
-        In particular, set the User-Agent header to identify the GX Agent and the correlation_id if provided.
+        In particular, set the User-Agent header to identify the GX Agent and the correlation_id as
+        Agent-Job-id if provided.
+
+        Note: the Agent-Job-Id header value will be set for all GX Cloud request until this method is
+        called again.
         """
         from great_expectations import __version__
         from great_expectations.core import http
@@ -307,7 +311,9 @@ class GXAgent:
             "0.19"  # using 0.19 instead of 1.0 to account for pre-releases
         ):
             # TODO: public API should be available in v1
-            LOGGER.info("Unable to set User-Agent header for requests to GX Cloud")
+            LOGGER.info(
+                f"Unable to set {HeaderName.USER_AGENT} or {HeaderName.AGENT_JOB_ID} header for requests to GX Cloud"
+            )
             return
 
         def _update_headers_agent_patch(
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -81,7 +81,7 @@ class GXAgent:
     def __init__(self: Self):
         agent_version: str = self._get_current_gx_agent_version()
         print(f"GX Agent version: {agent_version}")
-        print(f"Initializing the GX Agent version: {agent_version}.")
+        print("Initializing the GX Agent.")
         self._config = self._get_config()
         self._set_http_session_headers()
         print("Loading a DataContext - this might take a moment.")
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -82,7 +82,7 @@ class GXAgent:
         agent_version: str = self._get_current_gx_agent_version()
         print(f"Initializing the GX Agent version: {agent_version}.")
         self._config = self._get_config()
-        self._set_user_agent_header()
+        self._set_http_session_headers()
         print("Loading a DataContext - this might take a moment.")
 
         with warnings.catch_warnings():
@@ -153,6 +153,9 @@ class GXAgent:
             loop = asyncio.get_event_loop()
             loop.create_task(event_context.redeliver_message())
             return
+
+        # ensusre that great_expectations.http requests to GX Cloud include the job_id/correlation_id
+        self._set_http_session_headers(correlation_id=event_context.correlation_id)
 
         # send this message to a thread for processing
         self._current_task = self._executor.submit(self._handle_event, event_context=event_context)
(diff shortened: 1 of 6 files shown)


tests/agent/test_agent.py
```diff
@@ -230,3 +233,30 @@ def consume(queue: str, on_message: Callable[[EventContext], None]):
...
+    gx_agent_config: GXAgentConfig,
+):
+    """Ensure custom User-Agent header is set on GX Cloud api calls."""
+    base_url = gx_agent_config.gx_cloud_base_url
+    org_id = gx_agent_config.gx_cloud_organization_id
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            responses.GET,
+            f"{base_url}/organizations/{org_id}/data-context-configuration",
+            # TODO: move the elsewhere (use VCR'ish yaml?)
```
Kilo59, line 250:
```suggestion
```

great_expectations_cloud/agent/agent.py
```diff
@@ -40,6 +42,7 @@
 )
 
 if TYPE_CHECKING:
```
allensallinger, line 44: Why are all of these imports under this `if TYPE_CHECKING` block?

Kilo59: Because they are only being used for type annotations and aren't needed at runtime.

https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking

great_expectations_cloud/agent/agent.py
```diff
@@ -123,9 +128,9 @@ def _listen(self) -> None:
             if subscriber is not None:
                 subscriber.close()
 
-    def _get_current_gx_agent_version(self) -> str:
-        version: str = metadata_version(self._PYPI_GX_AGENT_PACKAGE_NAME)
-        print(f"GX Agent version: {version}")
+    @classmethod
+    def _get_current_gx_agent_version(cls) -> str:
+        version: str = metadata_version(cls._PYPI_GX_AGENT_PACKAGE_NAME)
         return version
```
Kilo59, line 134: @Shinnnyshinshin I moved the print side-effect from this method to the `__init__` where it is called.

great_expectations_cloud/agent/agent.py
```diff
@@ -77,8 +80,10 @@ class GXAgent:
 
     def __init__(self: Self):
         agent_version: str = self._get_current_gx_agent_version()
+        print(f"GX Agent version: {agent_version}")
```
JennyTee, line 83: Should we remove the `agent_version` from line 84 then so it's not printed twice?

great_expectations_cloud/agent/agent.py
```diff
@@ -150,6 +155,9 @@ def _handle_event_as_thread_enter(self, event_context: EventContext) -> None:
             loop.create_task(event_context.redeliver_message())
             return
 
+        # ensusre that great_expectations.http requests to GX Cloud include the job_id/correlation_id
```
JennyTee, line 158: super minor
```suggestion
        # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
```

great_expectations_cloud/agent/agent.py
```diff
@@ -287,6 +295,46 @@ def _update_status(self, job_id: str, status: JobStatus) -> None:
...
+            return
+
+        def _update_headers_agent_patch(
+            session: requests.Session, access_token: str
+        ) -> requests.Session:
+            headers = {
+                "Content-Type": "application/vnd.api+json",
+                "Authorization": f"Bearer {access_token}",
+                "Gx-Version": __version__,
+                HeaderName.USER_AGENT: f"{USER_AGENT_HEADER}/{cls._get_current_gx_agent_version()}",
```
JennyTee, line 327: I never knew about the user-agent header; this is cool

great_expectations_cloud/agent/agent.py
Kilo59: I agree

great_expectations_cloud/agent/agent.py
Kilo59: thank you!

great_expectations_cloud/agent/agent.py
```diff
@@ -123,9 +128,9 @@ def _listen(self) -> None:
             if subscriber is not None:
                 subscriber.close()
 
-    def _get_current_gx_agent_version(self) -> str:
-        version: str = metadata_version(self._PYPI_GX_AGENT_PACKAGE_NAME)
-        print(f"GX Agent version: {version}")
+    @classmethod
+    def _get_current_gx_agent_version(cls) -> str:
```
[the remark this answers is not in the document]
Kilo59: I have to call it with `self` because `__init__` is an instance method.
I can call a class method from an instance method but I can't call an instance method from a class method.

great_expectations_cloud/agent/agent.py
```diff
@@ -287,6 +295,46 @@ def _update_status(self, job_id: str, status: JobStatus) -> None:
...
+                "Gx-Version": __version__,
+                HeaderName.USER_AGENT: f"{USER_AGENT_HEADER}/{cls._get_current_gx_agent_version()}",
+            }
+            if correlation_id:
+                headers[HeaderName.AGENT_JOB_ID] = correlation_id
+            session.headers.update(headers)
+            return session
+
+        # TODO: this is relying on a private implementation detail
+        # use a public API once it is available
```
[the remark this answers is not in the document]
Kilo59: I just created one and linked it on Slack.

PR #100, opened 2024-01-25
update agent printout error message for clarity

JennyTee:
Prior to this change, users noted some confusion about the error message used in the GX Agent printout, since sometimes this error message happens when a job is completed with errors (as opposed to when a job fails to complete), like when testing Data Source config.

For that reason I updated the error message to indicate that the job completed with errors. Future enhancements could add a more-specific message, but this should be fine for now.

Before change:
image

After change:
image


anthonyburdi (approved):
Much clearer, thank you!

PR #101, opened 2024-01-26
Add `pyscopg2` as optional dependency

Kilo59:
Adding the `psycopg2` package as an optional dependency

https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary

Can be installed using the `postgres` extra
```
pip install 'great_expectations[postgres]'
```
Or with poetry

```
poetry install --with sql
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.21"
+version = "0.0.22"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 repository = "https://github.com/great-expectations/cloud"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -31,17 +31,21 @@ packaging = "^21.3"
...
+postgres = [
+    "sqlalchemy",
+    "psycopg2",
+]
 
 [tool.poetry.group.sql.dependencies]
 great-expectations = { version = "^0.18.4", extras = [
     "databricks",
     "sqlalchemy1",
-    "postgres",
```
Kilo59, line 39: Removed this because using this extra will install `psycopg2-binary`
https://github.com/great-expectations/great_expectations/blob/develop/reqs/requirements-dev-postgresql.txt

anthonyburdi (approved):
LGTM

PR #105, opened 2024-01-26
Configure Dependabot Version Updates

anthonyburdi:
Configure dependabot version updates. This will raise pull requests to keep our dependencies up to date.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-dependabotyml-file

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory

This PR should not bump the agent version.

PR #106, opened 2024-01-26
Revert "Configure Dependabot Version Updates"

anthonyburdi:
Reverts great-expectations/cloud#105

It looks like maybe this isn't ready for prime time yet. Getting a "invalid details" error using `poetry` as the `package-ecosystem` here: https://github.com/great-expectations/cloud/runs/20914663368
Even though poetry is listed on the list of supported package ecosystems: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,0 @@
-# Please see the documentation for all configuration options:
-# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
-
-version: 2
-updates:
-  - package-ecosystem: "poetry" # See documentation for possible values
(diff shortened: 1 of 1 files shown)


Kilo59:
>Tip: For package managers such as pipenv and poetry, you need to use the pip YAML value. For example, if you use poetry to manage your Python dependencies and want Dependabot to monitor your dependency manifest file for new versions, use package-ecosystem: "pip" in your dependabot.yml file.

Kilo59 (changes requested):
This is fixable, let fix rather than revert

anthonyburdi:
Thanks - did not see that tip.

PR #113, opened 2024-01-30
Add `yellow_tripdata.db` to Docker Image

Kilo59:
Adds `yellow_tripdata.db` (7MB) sqlite database to the docker image.

This ensures that every agent has access to some data they can connect to explore and run agent-powered interactions from the cloud.


examples/agent/data/yellow_tripdata.db
Kilo59, line 1: Okay I lied, it's closer to `7MB` 😄

allensallinger (approved):

PR #115, opened 2024-01-31
update config error message text; avoid printing stacktrace for config issues at startup

JennyTee:
Reason for change: Provide a better UX when starting the GX Agent by making config errors easier to read/understand. 

Note: I did not want to affect any other existing errors that currently provide a stacktrace (which is being surfaced to the GX Cloud UI via the GX Agent logs, at least for now, and is helpful for troubleshooting). In this PR I basically 1) update the error message text to be more readable and 2) add special handling for a config-related errors so that they are caught, logged, and do not print a messy stacktrace.

If people think this approach isn't worth the added code complexity, an alternative solution could be to just update the error message text to be more readable and keep the stacktrace that is shown in the first screenshot below.

This is admittedly a bit of an edge case since if users follow instructions, these config variables should always be set.

Before change:
image

After change:
image


anthonyburdi (changes requested):
Looks good other than a minor refactor request, and I think this probably warrants tests in `tests/agent/test_agent.py` to check that the correct error was raised with incorrect env vars & incorrect config.

great_expectations_cloud/agent/agent.py
```diff
@@ -282,8 +285,11 @@ def _get_config(cls) -> GXAgentConfig:
...
             )
         except pydantic.ValidationError as validation_err:
-            raise GXAgentError(
-                f"Missing or badly formed environment variable\n{validation_err.errors()}"
+            missing_variables = ", ".join(
+                [f"{validation_error['loc'][0]}" for validation_error in validation_err.errors()]
+            )
+            raise GXAgentConfigError(
+                f"Missing or badly formed environment variable(s). Make sure to set the following environment variable(s): {missing_variables}"
             ) from validation_err
```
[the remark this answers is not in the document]
JennyTee: good call, thanks - I created a logging_util.py file that handles this now. And also unit tests for that file.

JennyTee:
> Looks good other than a minor refactor request, and I think this probably warrants tests in `tests/agent/test_agent.py` to check that the correct error was raised with incorrect env vars & incorrect config.

Thanks - added tests to `test_agent.py` for the invalid config error!

tests/agent/test_agent.py
```diff
@@ -39,7 +39,53 @@ def set_required_env_vars(monkeypatch, org_id, token):
...
+    monkeypatch.setattr(os, "environ", env_vars)
+
+
+@pytest.fixture
+def set_required_env_vars_missing_org_id(monkeypatch, token):
+    env_vars = {
+        "GX_CLOUD_ORGANIZATION_ID": None,
+        "GX_CLOUD_ACCESS_TOKEN": token,
+    }
+    monkeypatch.setattr(os, "environ", env_vars)
```
Kilo59, line 60: You should use `monkeypatch.setenv` for this kind of thing.

https://docs.pytest.org/en/latest/how-to/monkeypatch.html#monkeypatching-environment-variables

tests/agent/test_agent.py
```diff
@@ -33,7 +33,7 @@
     from tests.agent.conftest import DataContextConfigTD
 
 
-@pytest.fixture(autouse=True)
```
JennyTee, line 36: (Since I added a test with bad config, we no longer want to autouse this fixture)

tests/test_logging_util.py
```diff
@@ -0,0 +1,54 @@
...
+    string_one: str
+    string_two: str
+
+
+@pytest.fixture
+def validation_error_missing_string_one():
+    try:
+        LoggingUtilTestModel(string_one=None, string_two="i am string two")
+    except ValidationError as validation_error:
+        return validation_error
```
JennyTee, line 22: In this method I'm forcing a validation error from pydantic. This way if the pydantic model structure for their ValidationErrors ever changes, this test will fail and we can rewrite the way we pull the missing attribute names from the error.

great_expectations_cloud/agent/agent.py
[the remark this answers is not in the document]
anthonyburdi: Thank you! I was originally thinking of a private method in `GXAgent` but this works too! What do you think about moving the function into `great_expectations_cloud/agent/agent.py` as a module level function instead of a util file? I tend to shy away from util files since they can become cluttered and disconnected from the context where they are used. 
Here is a pretty good [writeup on util files](https://softwareengineering.stackexchange.com/a/422932) with some add'l rationale.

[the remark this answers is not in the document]
JennyTee: Thanks; saving that stackoverflow link :) 

As discussed, I moved this code to the config.py file since it's a pretty print method that's specifically used for printing config validation errors. (Also renamed it.)

There are a couple of other places we can use this method too where we are checking config; I'll add those in another PR.

great_expectations_cloud/agent/config.py
```diff
@@ -9,3 +9,11 @@ class GxAgentEnvVars(BaseSettings):  # type: ignore[misc] # BaseSettings is has
     gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL
     gx_cloud_organization_id: str
     gx_cloud_access_token: str
+
+
+def generate_config_validation_error_text(validation_error: ValidationError) -> str:
```
anthonyburdi, line 14: Nice! Thanks for moving this! I like the method rename too.

tests/agent/test_agent.py
JennyTee: nice! ty - updated these tests to use `monkeypatch.setenv` instead

anthonyburdi (approved):
Thanks for the changes! Looks great!

PR #116, opened 2024-01-31
Fix missing `Agent-Job-Id` for `CheckpointRun` events 

Kilo59:
The `AgentJobId` header was not being set for the `RunCheckpointEvent`/`RunCheckpointAction` but is for other events like `DraftDatasourceConfigEvent `

- Initial implementation https://github.com/great-expectations/cloud/pull/97

This will ensure that `Agent-Job-Id` is set for any GX Cloud API call that originates from core OSS code.

### Notes

`_set_http_session_headers()` is modifying global state (it's monkeypatching a method in the OSS code), and can be called with or without a `correlation_id`.
https://github.com/great-expectations/cloud/blob/a3fee2fbcd5c041a66ebbfbf3e545450cc49ed6c/great_expectations_cloud/agent/agent.py#L166-L167

~My initial guess was that `_set_http_session_headers()` was being called without the `correlation_id` and this was causing the `Agent-Job-Id` to be unset, but after adding more logging this doesn't appear to be the case. `_set_http_session_headers()` is only called once without a `correlation_id` when the agent is first setting up the context, after that it always has a `correlation_id` provided.~

The issue was that the stores use a different session that is created before we have a `correlation_id`.
The fix is to update `_set_http_session_headers()` to apply the `correlation_id`/`Agent-Job-Id` header to each `GXCloudStoreBackend` session.

https://github.com/great-expectations/great_expectations/blob/91240de2862a33b7e7016c274b31e6693405cc78/great_expectations/data_context/store/gx_cloud_store_backend.py#L207-L209

`DraftDatasourceConfigEvent` is not using a `GXCloudStoreBackend` to make these HTTP calls and so didn't suffer from this problem.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -164,7 +164,6 @@ class GXAgent:
             return
 
         # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
-        assert event_context.correlation_id  # TODO: remove this
         self._set_http_session_headers(correlation_id=event_context.correlation_id)
 
         # send this message to a thread for processing
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -314,6 +314,7 @@ class GXAgent:
         """
         from great_expectations import __version__
         from great_expectations.core import http
+        from great_expectations.data_context.store.gx_cloud_store_backend import GXCloudStoreBackend
 
         if Version(__version__) > Version(
             "0.19"  # using 0.19 instead of 1.0 to account for pre-releases
@@ -332,7 +333,9 @@ class GXAgent:
         if correlation_id:
             # OSS doesn't use the same session for all requests, so we need to set the header for each store
             for store in self._context.stores.values():
-                store._store_backend._session.headers[HeaderName.AGENT_JOB_ID] = correlation_id
+                backend = store._store_backend
+                if isinstance(backend, GXCloudStoreBackend):
+                    backend._session.headers[HeaderName.AGENT_JOB_ID] = correlation_id
 
         def _update_headers_agent_patch(
             session: requests.Session, access_token: str
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -329,10 +329,17 @@ class GXAgent:
             f"Setting session headers for GX Cloud. {HeaderName.USER_AGENT}:{agent_version} {HeaderName.AGENT_JOB_ID}:{correlation_id}"
         )
 
+        if correlation_id:
+            # OSS doesn't use the same session for all requests, so we need to set the header for each store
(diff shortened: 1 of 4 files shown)


tests/agent/test_agent.py
```diff
@@ -294,31 +319,34 @@ def test_correlation_id_header(
...
+            f"{base_url}organizations/{org_id}/checkpoints/{checkpoint_id}",
+            json={"data": {}},
             # match will fail if correlation-id header is not set
             match=[
-                responses.matchers.header_matcher({HeaderName.AGENT_JOB_ID: str(agent_job_id_2)})
+                responses.matchers.header_matcher(
+                    {
+                        HeaderName.USER_AGENT: f"{USER_AGENT_HEADER}/{GXAgent._get_current_gx_agent_version()}",  # TODO: remove
+                        HeaderName.AGENT_JOB_ID: agent_job_ids[2],
+                    }
```
Kilo59, line 348: Note: `User-Agent` is being set but `Agent-Job-Id` is not.
But this only happens for `RunCheckpointEvent`; for `DraftDatasourceConfigEvent`, it's set without issue.

great_expectations_cloud/agent/agent.py
```diff
@@ -85,8 +85,8 @@ def __init__(self: Self):
         print(f"GX Agent version: {agent_version}")
         print(f"Great Expectations version: {great_expectations_version}")
         print("Initializing the GX Agent.")
-        self._config = self._get_config()
         self._set_http_session_headers()
+        self._config = self._get_config()
```
Kilo59, line 89: This is very minor and wasn't the cause of this bug, but ensure that the `/agent-session` call includes the custom `User-Agent` header.

great_expectations_cloud/agent/agent.py
```diff
@@ -164,6 +164,7 @@ def _handle_event_as_thread_enter(self, event_context: EventContext) -> None:
             return
 
         # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
+        assert event_context.correlation_id  # TODO: remove this
```
Kilo59, line 167:
```suggestion
```

great_expectations_cloud/agent/agent.py
```diff
@@ -324,14 +324,29 @@ def _set_http_session_headers(cls, correlation_id: str | None = None) -> None:
             )
             return
 
+        agent_version = self._get_current_gx_agent_version()
+        LOGGER.info(
+            f"Setting session headers for GX Cloud. {HeaderName.USER_AGENT}:{agent_version} {HeaderName.AGENT_JOB_ID}:{correlation_id}"
+        )
```
Kilo59, line 330:
```suggestion
        LOGGER.debug(
            f"Setting session headers for GX Cloud. {HeaderName.USER_AGENT}:{agent_version} {HeaderName.AGENT_JOB_ID}:{correlation_id}"
        )
```

PR #117, opened 2024-02-01
[BUGFIX] Do on try to reconnect on authentication errors

TrangPham:
It doesn't make sense to try to reconnect on these errors: StreamLostError, ConnectionClosedByBroker, AuthenticationError

AuthenticationError would not resolve via retrying

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, develop, --branch, main]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.1.13"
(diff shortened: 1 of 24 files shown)


allensallinger (approved):
The changes look good, bump the version and we should be good to go!

wookasz (commented):
@TrangPham agreed that authetnication error won't make sense to retry as credentials won't just become correct on retry, however, why would we not retry the others.

SteramLostError seems like it can happen if heartbeats fail, reconnecting should resolve the issue?
Can ConnectionClosedByBroker happen when broker restarts?

for AuthetnicationError, we should try to handle this by making another call to the mercury `agent-sessions` endpoint to get a new set of credentials

PR #120, opened 2024-02-02
Bugfix - `No module named 'psycopg2'`

Kilo59:
The source install of `pyscopg2` is not working for our docker image and requires some additional work to fix.

Reverting to using `pyscopg2-binary` as this seems to work without issues. Comments from one of the `pyscopg2` developers suggest that using this for our agent won't actually be a problem as our agent will only be connecting to postgres in a single thread.

>The use of the -binary packages in production is discouraged because in the past they proved unreliable in multithread environments. This might have been fixed in more recent versions but I have never managed to reproduce the failure.

## References

- https://github.com/great-expectations/cloud/pull/101
- https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary
- https://github.com/psycopg/psycopg2/issues/1283#issuecomment-850333031
- https://stackoverflow.com/questions/12906351/importerror-no-module-named-psycopg2

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -73,9 +73,24 @@ def lint(ctx: Context, check: bool = False, unsafe_fixes: bool = False) -> None:
     ctx.run(" ".join(cmds), echo=True, pty=True)
 
 
-@invoke.task
-def docker(ctx: Context, check: bool = False, tag: str = "greatexpectations/agent:develop") -> None:
-    """Lint Dockerfile using hadolint tool"""
+@invoke.task(
+    aliases=["build"],
+    help={
+        "check": "Lint Dockerfile using hadolint tool",
+        "run": "Run the Docker container. Inject .env file",
+    },
+)
+def docker(
+    ctx: Context,
+    check: bool = False,
+    run: bool = False,
+    tag: str = "greatexpectations/agent:develop",
+) -> None:
+    """
+    Docker tasks
+
+    If no options are provided, the default behavior is to build the Docker image.
+    """
     if check:
         cmds = [
             "docker",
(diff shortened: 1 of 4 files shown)


tasks.py
```diff
@@ -118,22 +136,6 @@ def deps(ctx: Context) -> None:
...
-        "docker",
-        "buildx",
-        "build",
-        "-t",
-        "greatexpectations/agent",
-        "-f",
-        DOCKERFILE_PATH,
-        ".",
-    ]
-    ctx.run(" ".join(cmds), echo=True, pty=True)
```
Kilo59, line 134: This seems like a duplicate of the `docker` task.
What's the difference between `buildx build` and `build`?

Kilo59:
Note I did try installing both `libpq-dev` and `openssl` in the Dockerfile before giving up.

I made no attempt to "link" (symlink) `libpq` as is suggested in the Mac OS instructions here.
https://stackoverflow.com/a/67757077/6304433

tasks.py
allensallinger: https://docs.docker.com/build/architecture/

Buildx is the new build tool that docker uses. Under the hood newer versions of docker run buildx when you run docker build.

allensallinger (approved):
LGTM, one comment but it should be good.

tasks.py
```diff
@@ -92,8 +107,11 @@ def docker(ctx: Context, check: bool = False, tag: str = "greatexpectations/agen
             "<",
             DOCKERFILE_PATH,
         ]
+    elif run:
+        cmds = ["docker", "run", "--env-file .env", "--rm", "-t", tag]
```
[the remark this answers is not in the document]
Kilo59: Whoops it was on auto-merge.
I can followup with a change.

PR #127, opened 2024-02-07
Test connection before running checkpoint

JennyTee:
In order to provide the UI with an error message that it knows how to handle, the agent needs to test the Data Source connection prior to actually running the validation.

This PR handles having the Agent test the connection. The changes here depend on the changes in https://github.com/greatexpectationslabs/mercury/pull/2772.


great_expectations_cloud/agent/models.py
```diff
@@ -38,6 +38,7 @@ class RunMissingnessDataAssistantEvent(RunDataAssistantEvent):
 
 class RunCheckpointEvent(EventBase):
     type: Literal["run_checkpoint_request"] = "run_checkpoint_request"
+    datasource_name: Union[str, None] = None
```
JennyTee, line 41: Nullability of param discussed here: https://github.com/greatexpectationslabs/mercury/pull/2772#discussion_r1481944407

great_expectations_cloud/agent/actions/run_checkpoint.py
```diff
@@ -15,10 +15,16 @@
 class RunCheckpointAction(AgentAction[RunCheckpointEvent]):
     @override
     def run(self, event: RunCheckpointEvent, id: str) -> ActionResult:
+        # TODO: move connection testing into OSS; there isn't really a reason it can't be done there
+        for datasource_name in event.datasource_names:
+            datasource = self._context.get_datasource(datasource_name)
+            datasource.test_connection(test_assets=True)  # raises `TestConnectionError` on failure
```
Kilo59, line 21: I would suggest only testing the connection for the asset that the check is being run against.
A problem with asset `A` shouldn't affect my ability to validate asset `B`.

Every asset type also has a `test_connection()` method that can called.
```python
# something like this
        for datasource_name in event.datasource_names:
            datasource = self._context.get_datasource(datasource_name)
            datasource.test_connection(test_assets=False)
            for asset_name in checkpoint_assets:
                asset = datasource.get_asset(asset_name)
                asset.test_connection()
```

tests/agent/actions/checkpoints/test_run_checkpoint_action.py
```diff
@@ -97,3 +105,23 @@ def test_run_checkpoint_action_with_splitter_options_returns_action_result(
             type="SuiteValidationResult",
         ),
     ]
+
+
+def test_run_checkpoint_action_raises_on_test_connection_failure(
+    context, checkpoint_id, datasource_names
+):
+    mock_datasource = MagicMock()
```
Kilo59, line 113:
```suggestion
    mock_datasource = MagicMock(spec=Datasource)
```
MagicMock without a spec is a dangerous footgun.
It will recursively create mocks for any attribute that you access.

Kilo59 (changes requested):
If possible I think we should only be testing the assets that the checkpoint is being run against.

Also minor suggestion about the use of `MagicMock`.

great_expectations_cloud/agent/actions/run_checkpoint.py
JennyTee: makes sense--thanks for the explanation; will update!

JennyTee: ok, this PR has been updated - I also updated the [corresponding mercury PR](https://github.com/greatexpectationslabs/mercury/pull/2772)

great_expectations_cloud/agent/models.py
```diff
@@ -38,6 +38,7 @@ class RunMissingnessDataAssistantEvent(RunDataAssistantEvent):
 
 class RunCheckpointEvent(EventBase):
     type: Literal["run_checkpoint_request"] = "run_checkpoint_request"
+    datasource_names_to_asset_names: dict
```
Kilo59, line 41:
```suggestion
    datasource_names_to_asset_names: dict[str, str]
```

Kilo59 (approved):
Thanks for this.

One last non-blocking suggestion is to think about if we want to fail-fast if a single asset cannot connect or do we want to accumulate all asset connection failures and then return the details of all those errors at the end?
I think I have a very slight preference for failing-fast which is what you have now, but both have their benefits and costs.

great_expectations_cloud/agent/models.py
JennyTee: I think this syntax only works with Python 3.9 and up. I had tried it before (and forgot and just tried it again with the suggestion above :) and the Python 3.8 tests fail with `TypeError: 'type' object is not subscriptable`

Not sure if there is another way around this.

JennyTee:
> Thanks for this.
> 
> One last non-blocking suggestion is to think about if we want to fail-fast if a single asset cannot connect or do we want to accumulate all asset connection failures and then return the details of all those errors at the end? I think I have a very slight preference for failing-fast which is what you have now, but both have their benefits and costs.

Thanks so much for the detailed reviews, @Kilo59! I talked to @anthonyburdi last week about failing fast and we agree with you that failing fast is likely preferable. I'll mention it to Anna too so that she is aware. Will create a separate ticket if she thinks we should look into other options.

great_expectations_cloud/agent/models.py
Kilo59:
```python
from typing import Dict, Set

...

datasource_names_to_asset_names: Dict[str, Set[str]]
```

PR #129, opened 2024-02-09
Fail CDM job if any metric fails (after storing results)

anthonyburdi:
If any metric fails during a CDM run, fail the agent job by raising an exception. Before raising the exception, store the results so that a user can view incomplete results if available.


great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
```diff
@@ -44,10 +45,18 @@ def run(self, event: RunColumnDescriptiveMetricsEvent, id: str) -> ActionResult:
 
         metric_run_id = self._metric_repository.add_metric_run(metric_run)
 
+        # Note: This exception is raised after the metric run is added to the repository so that
+        # the user can still access any computed metrics even if one of the metrics fails.
```
JennyTee, line 49: ty for this comment!

JennyTee (approved):
✨

PR #130, opened 2024-02-12
Update event models to allow extra properties to ensure backwards compatibility

JennyTee:
In #127, we are adding a property to an existing Event model. This change will be merged to both `mercury` and `cloud`, but we currently don't allow for extra properties on event models...so if #127 is merged and users do not immediately upgrade their GX Agent version, the RunCheckpointEvent type is not recognized and therefore cannot be processed, which breaks running checkpoints/validations.

To ensure backwards compatibility, we need to make sure that the Agent code can recognize events even if they have additional properties; hence this change.

I'm open to doing this another way if there are suggestions.

### DIFF
--- a/great_expectations_cloud/agent/models.py
+++ b/great_expectations_cloud/agent/models.py
@@ -10,7 +10,7 @@ from typing_extensions import Annotated
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
     class Config:
-        extra: str = Extra.allow
+        extra: str = Extra.ignore
 
 
 class EventBase(AgentBaseModel):
--- a/great_expectations_cloud/agent/models.py
+++ b/great_expectations_cloud/agent/models.py
@@ -10,7 +10,7 @@ from typing_extensions import Annotated
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
     class Config:
-        extra: str = Extra.forbid
+        extra: str = Extra.allow
 
 
 class EventBase(AgentBaseModel):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.28"
+version = "0.0.29"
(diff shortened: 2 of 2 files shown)


anthonyburdi (approved):
I think this is reasonable to do for now. An alternative is to version events, but that is a much bigger lift so we should consider that separately. FYI I think you'll need to bump the version again after merging `main` since we are at 0.0.29 already.

JennyTee:
> I think this is reasonable to do for now. An alternative is to version events, but that is a much bigger lift so we should consider that separately. FYI I think you'll need to bump the version again after merging `main` since we are at 0.0.29 already.

ty @anthonyburdi! I bumped the version again.

fwiw I don't believe we currently have any tests set up to check for backwards compatibility with different versions of the Agent. Would have been easy for me to just have pushed the update to RunCheckpointEvent without manually checking for backwards compatibility. 😬 Probably something for us to consider moving forward.

great_expectations_cloud/agent/models.py
```diff
@@ -10,7 +10,7 @@
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
     class Config:
-        extra: str = Extra.forbid
+        extra: str = Extra.allow
```
Kilo59, line 13: Can we change this to just ignore extra fields?
```suggestion
        extra: str = Extra.ignore
```
Going from `forbid` to `allow` seems drastic.

If we start `allowing` extra fields the schema loses a lot of its value.

Kilo59:
I would suggest creating a new event or using a FF to toggle between the new and old version of the Event.

Kilo59 (changes requested):
I think we need to think about this more.

Even with this change users will still need to update their agent just as they would without the schema loosening.

Kilo59 (approved):

JennyTee:
> I think we need to think about this more.
> 
> Even with this change users will still need to update their agent just as they would without the schema loosening.

Sounds good. Created https://greatexpectations.atlassian.net/browse/ZEL-498 to track this; will plan on merging with `extra.ignore` per Slack convo to unblock current feature work

PR #136, opened 2024-02-14
Update ruff to 0.2.1

Kilo59:
Updating to `ruff` 0.21`

https://astral.sh/blog/ruff-v0.2.0#migrating-to-v020


pyproject.toml
```diff
@@ -94,7 +94,7 @@ line-length = 100
 target-version = "py38"
 # TODO: change this back to black default of 88 if desired
 line-length = 100
-select = [
+lint.select = [
```
Kilo59, line 97: This is part of the migration to `0.21`; now that `ruff` is also a formatter, they have namespace the linter settings to distinguish it from the formatter settings.

rreinoldsc (approved):

PR #137, opened 2024-02-15
ZEL-503: Only test connection for profiled data asset

JennyTee:
We've been testing the connection for all of the data assets in a datasource. When users run CDM, we only need to test the connection for the data asset that is being profiled.

With the changes in this PR, here is what the error messages look like for a nonexistent table and a Data Source with bad creds:

image


anthonyburdi (approved):
LGTM!

PR #141, opened 2024-02-23
[BUGFIX] gx agent pika connection errors when 2 agents are started at the same time

TrangPham:
When there are two or more agents attempting to connect two the queue, each generates a password and this creates a race condition where the first can connect but when another tries with the password they generated it would have auth error.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -142,6 +142,11 @@ class GXAgent:
             print("Received request to shutdown.")
         except (SubscriberError, ClientError):
             print("Connection to GX Cloud has encountered an error.")
+        except AuthenticationError:
+            # Retry with new credentials
+            self._config = self._get_config()
+            # Raise to use the retry decorator to handle the retry logic
+            raise
         finally:
             if subscriber is not None:
                 subscriber.close()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.34.dev0"
+version = "0.0.34.0"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 repository = "https://github.com/great-expectations/cloud"
@@ -29,13 +29,8 @@ packaging = "^21.3"
 tenacity = "^8.2.3"
 
 [tool.poetry.extras]
-snowflake = [
-    "snowflake-sqlalchemy",
-]
-postgres = [
-    "sqlalchemy",
-    "psycopg2-binary",
-]
+snowflake = ["snowflake-sqlalchemy"]
+postgres = ["sqlalchemy", "psycopg2-binary"]
 
 [tool.poetry.group.sql.dependencies]
 great-expectations = { version = "^0.18.4", extras = [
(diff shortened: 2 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -114,6 +116,16 @@ def run(self) -> None:
         self._listen()
         print("Connection to GX Cloud has been closed.")
 
+    # ZEL-505: A race condition can occur if two or more agents are started at the same time
+    #          due to the generation of passwords for rabbitMQ queues. This can be mitigated
+    #          by adding a delay and retrying the connection. Retrying with new credentials
+    #          requires calling get_config again, which handles the password generation.
+    @retry(
+        retry=retry_if_exception_type(AuthenticationError),
+        wait=wait_exponential(multiplier=1, min=4, max=10),
```
rreinoldsc, line 125: I'd set starting wait to 1s, unless strong reason it should be 4s
```suggestion
        wait=wait_exponential(multiplier=1, min=1, max=10),
```

pyproject.toml
```diff
@@ -26,6 +26,8 @@ pika = "^1.3.1"
 orjson = "^3.9.7, !=3.9.10" # TODO: remove inequality once dep resolution issue is resolved
 # relying on packaging in agent code so declaring it explicitly here
 packaging = "^21.3"
+tenacity = "^8.2.3"
+ruff = "0.2.1"
```
rreinoldsc, line 30:
```suggestion
```

Already in [tool.poetry.group.dev.dependencies]

great_expectations_cloud/agent/agent.py
```diff
@@ -114,6 +116,16 @@ def run(self) -> None:
         self._listen()
         print("Connection to GX Cloud has been closed.")
 
+    # ZEL-505: A race condition can occur if two or more agents are started at the same time
+    #          due to the generation of passwords for rabbitMQ queues. This can be mitigated
+    #          by adding a delay and retrying the connection. Retrying with new credentials
```
rreinoldsc, line 121: Can we support two valid sets of creds?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "0.0.33"
+version = "0.0.34.dev0"
```
anthonyburdi, line 3: Are we using dev releases?

anthonyburdi (commented):
Looks good pending Robby's question and clarification on dev vs non dev release version. I think in this case we do want a new non-dev version so that the image on dockerhub is updated.

great_expectations_cloud/agent/agent.py
wookasz: the login happens only once and lasts the duration of the session. I.e. a connection is not terminated when the credentials change

great_expectations_cloud/agent/agent.py
```diff
@@ -114,6 +116,16 @@ def run(self) -> None:
...
+    #          requires calling get_config again, which handles the password generation.
+    @retry(
+        retry=retry_if_exception_type(AuthenticationError),
+        wait=wait_exponential(multiplier=1, min=1, max=10),
+        stop=stop_after_attempt(3),
+        after=after_log(LOGGER, logging.DEBUG),
+    )
     def _listen(self) -> None:
         """Manage connection lifecycle."""
         subscriber = None
```
wookasz, line 131: Do you need to call `self._config = self._get_config()` in here on retry?

We may have the wrong credentials

great_expectations_cloud/agent/agent.py
rreinoldsc: oh I see. the race condition window is narrower than I was imaging. Thanks for clarifying

rreinoldsc (approved):

great_expectations_cloud/agent/agent.py
```diff
@@ -130,6 +142,11 @@ def _listen(self) -> None:
             print("Received request to shutdown.")
         except (SubscriberError, ClientError):
             print("Connection to GX Cloud has encountered an error.")
+        except AuthenticationError:
+            # Retry with new credentials
+            self._config = self._get_config()
```
wookasz, line 147: > I thought that _listen calls it.

Maybe it should? The way this works right now, we will call `_get_config` immediately after the auth error which may not be what we want. Instead, we may want to wait to fetch/set new credentials. Maybe move `_get_config` out of `_init` and put it at the start of `_listen`?

wookasz (approved):

PR #144, opened 2024-02-26
add unknown action

rreinoldsc:
If:
- new event type is received on agent, or
- an existing event type with unrecognized params

the UnknownEvent causes an UnknownAction, which reports back job status as completed=False, and a stack trace message


great_expectations_cloud/agent/agent.py
```diff
@@ -209,14 +208,24 @@ def _handle_event_as_thread_exit(
...
         if error is None:
             result: ActionResult = future.result()
-            status = JobCompleted(
-                success=True,
-                created_resources=result.created_resources,
-            )
-            print(f"Completed job: {event_context.event.type} ({event_context.correlation_id})")
+
+            # TODO This result may have an UnknownEvent error
+            if result.event is UnknownEvent:
```
rreinoldsc, line 216: this is the crux of changes

great_expectations_cloud/agent/agent.py
```diff
@@ -209,14 +208,25 @@ def _handle_event_as_thread_exit(
...
+                status = JobCompleted(
+                    success=False,
```
rreinoldsc, line 218: The data model is below, how should I convey its an unknown event @anthonyburdi ?
```
class JobCompleted(AgentBaseModel):
    status: Literal["completed"] = "completed"
    success: bool
    created_resources: Sequence[CreatedResource] = []
    error_stack_trace: Union[str, None] = None
```

great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
```diff
@@ -12,25 +22,22 @@
...
     ):
         super().__init__(context=context)
-        self._metric_repository = metric_repository
-        self._batch_inspector = batch_inspector
+        metric_retrievers: list[MetricRetriever] = [
+            ColumnDescriptiveMetricsMetricRetriever(self._context)
+        ]
+        cloud_data_store = CloudDataStore(self._context)
+        self._metric_repository = MetricRepository(data_store=cloud_data_store)
+        self._batch_inspector = BatchInspector(self._context, metric_retrievers)
```
Kilo59, line 41: Why is this changing?

great_expectations_cloud/agent/models.py
```diff
@@ -15,6 +27,7 @@ class Config:
 
 class EventBase(AgentBaseModel):
     type: str
+    action: ClassVar[AgentAction]
```
Kilo59, line 30:
```suggestion
    action: ClassVar[type[AgentAction]]
```

great_expectations_cloud/agent/agent.py
Kilo59: Previously an unknown event would be re-enqueued so that it potentially could be processed by a different agent that does know how to handle it.
This looks like it would break flow.

Kilo59:
I don't think you should mix refactoring with feature/behavior changes.
Can you breakup this PR into a refactoring PR and a behavior change PR?

great_expectations_cloud/agent/models.py
rreinoldsc: gosh, ty, I was struggling

PR #147, opened 2024-02-28
F/zel 483/zel 510/prototype compatibility

anthonyburdi:


great_expectations_cloud/agent/actions/compatibility.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def run_check_datasource_config(self) -> None:
+        ...
+
+
+_VERSION_RUNNERS = {}
+
+
+def register_version_runner(version: str, runner: VersionRunner) -> None:
+    _VERSION_RUNNERS[version] = runner
```
billdirks, line 23: nit: I'd check version isn't already present and throw if it is so we don't accidentally overwrite.

anthonyburdi: Good call, modified to:
```
def _register_version_runner(version: str, runner: _VersionRunner) -> None:
    if version in _VERSION_RUNNERS:
        raise ValueError(f"Version {version} already registered")
    _VERSION_RUNNERS[version] = runner
```
in https://github.dev/great-expectations/cloud/pull/148/files

I was also considering looking into registering when instantiating e.g. overriding `__new__`.

great_expectations_cloud/agent/actions/compatibility.py
```diff
@@ -0,0 +1,100 @@
...
+class V0Runner(VersionRunner):
+    @override
+    def run_checkpoint(self) -> None:
+        # ***** THIS IS WHERE THE 0.18 SPECIFIC CODE WOULD GO, MOVED FROM THE ACTION *****
+        print("Running run_checkpoint major version 0 (0.18)")
+
+    @override
+    def run_check_datasource_config(self) -> None:
+        # ***** THIS IS WHERE THE 0.18 SPECIFIC CODE WOULD GO, MOVED FROM THE ACTION *****
+        print("Running run_check_datasource_config major version 0 (0.18)")
```
Kilo59, line 35: Do we expect that the differences here are just schema/model/config changes or also changes to the workflow?

anthonyburdi: Scratch that - changing registration to be action/event/gx_version instead of the guts of the action as in #148.

anthonyburdi:
Closed in favor of https://github.com/great-expectations/cloud/pull/150

PR #149, opened 2024-02-28
Refactor CDM action

anthonyburdi:
Refactor the CDM action to match the signature of the parent action (context-only init). Provide defaults for the existing params.
Dev release since this is only a refactor that does not change functionality.

Thanks @billdirks for the catch!

### DIFF
--- a/great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
+++ b/great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
@@ -24,9 +24,6 @@ from great_expectations_cloud.agent.models import (
 
 if TYPE_CHECKING:
     from great_expectations.data_context import CloudDataContext
-    from great_expectations.experimental.metric_repository.metric_retriever import (
-        MetricRetriever,
-    )
     from great_expectations.experimental.metric_repository.metrics import MetricRun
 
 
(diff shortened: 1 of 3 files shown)


great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
```diff
@@ -25,12 +34,16 @@ class ColumnDescriptiveMetricsAction(AgentAction[RunColumnDescriptiveMetricsEven
...
         context: CloudDataContext,
-        metric_repository: MetricRepository,
-        batch_inspector: BatchInspector,
+        metric_repository: MetricRepository | None = None,
+        batch_inspector: BatchInspector | None = None,
     ):
         super().__init__(context=context)
-        self._metric_repository = metric_repository
-        self._batch_inspector = batch_inspector
+        metric_retrievers: list[MetricRetriever] = [
```
billdirks, line 41: This is only used when defining `self._batch_inspector` so, to prevent creating it when you don't need it, you could delete this and change line 46 to:
```
self._batch_inspector = batch_inspector or BatchInspector(context, ColumnDescriptiveMetricsMetricRetriever(self._context))
```

billdirks (approved):
One small comment but otherwise LGTM!

great_expectations_cloud/agent/actions/run_column_descriptive_metrics_action.py
anthonyburdi: Good call, same thing with `CloudDataStore`. Fixed!

PR #150, opened 2024-02-29
Refactor event handler to use event + action registry based on GX Core version

anthonyburdi:
This PR adds a registration mechanism to run the correct action based on the GX Core version installed with the agent.

This approach helps to prepare for catching errors for the Error Taxonomy work by allowing different implementations based on the GX Core version to be worked on in parallel (v0, v1). A side benefit will be to help enable the upcoming v1 compatibility work to be done in parallel with continuing to support v0 actions until the agent is upgraded fully to v1.

This PR is marked with a `dev` release version since it is mainly a refactor and does not change functionality.


great_expectations_cloud/agent/event_handler.py
```diff
@@ -3,37 +3,39 @@
...
+#         RunCheckpointEvent.__name__: RunCheckpointAction,
+#         RunColumnDescriptiveMetricsEvent.__name__: ColumnDescriptiveMetricsAction,
+#         DraftDatasourceConfigEvent.__name__: DraftDatasourceConfigAction,
+#     },
+#     "1": {
+#         # No events implemented yet for GX Core v1
+#         # Suggested naming convention, add version suffix "V1" e.g.: RunCheckpointEventV1
+#     },
+# }
+_EVENT_ACTION_MAP = {}
```
billdirks, line 36: You could type this `Dict[str, Dict[str, AgentAction]]`

billdirks (commented):
Overall the strategy looks good. I left a few comments. Tests should be added before merge.

Kilo59:
@rreinoldsc had a version of this that simply attached the `Action` to the `Event` itself and as a result the code was much simpler and there was no need for a `registry`

anthonyburdi:
> @rreinoldsc had a version of this that simply attached the `Action` class to the `Event` itself and as a result the code was much simpler and there was no need for a `registry`
> 
> https://github.com/great-expectations/cloud/pull/144/files/6be047f7599b3d3b0dfb8f05aca4a21a183e7b60#diff-3f202af1d34cf6ad91fad0d96f6ed7199ca4881d74d9747158cbe8f203425ae6

Love it. Part of the rationale though for the registry is that we can work on different actions for v0 vs v1 (major GX Core version). We can remove the registry after transitioning to v1. Unless I'm missing something in the implementation you linked. cc @rreinoldsc @billdirks

billdirks:
The generalization having a registry allows us is to associate more than 1 action with an event.

In the use case here, the same event will need to be processed by different code depending on the OSS version. If we attach the action to the event, then we will need to either update a particular action with new version code when we want to add a new version, or change what we attach to the event so it can route us to the correct action. The first strategy breaks the open/closed principle (the O in SOLID). That is the `run` method for an action runs code conditional on the OSS version. The second strategy is some type of registry (or conditional like the current event handler code). Having a registry, like the one in this PR, means when I want to extend which version is supported, we implement a new action and register it so all code is localized to the new action file.

great_expectations_cloud/agent/event_handler.py
```diff
@@ -3,37 +3,39 @@
...
+#         RunCheckpointEvent.__name__: RunCheckpointAction,
+#         RunColumnDescriptiveMetricsEvent.__name__: ColumnDescriptiveMetricsAction,
+#         DraftDatasourceConfigEvent.__name__: DraftDatasourceConfigAction,
+#     },
+#     "1": {
+#         # No events implemented yet for GX Core v1
+#         # Suggested naming convention, add version suffix "V1" e.g.: RunCheckpointEventV1
+#     },
+# }
+_EVENT_ACTION_MAP: dict[str, dict[str, type[AgentAction[Any]]]] = {}
```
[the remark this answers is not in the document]
anthonyburdi: When I do that I am getting a mypy error:
image

great_expectations_cloud/agent/event_handler.py
```diff
@@ -75,3 +66,32 @@ def handle_event(self, event: Event, id: str) -> ActionResult:
...
+
+
+def register_event_action(
+    version: str, event_type: type[Event], action_class: type[AgentAction[Any]]
+) -> None:
+    """Register an event type to an action class."""
+    version_in_map = (
+        _EVENT_ACTION_MAP.get(version) is not None
+    )  # check for None explicitly since version can be empty
+    if version_in_map and (event_type in _EVENT_ACTION_MAP[version]):
```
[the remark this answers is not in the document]
anthonyburdi: I was trying to raise before we modify `_EVENT_ACTION_MAP`. But if we do raise then it is moot what `_EVENT_ACTION_MAP` is since the program would die because we don't catch the exception raised in `register_event_action()`. I'll simplify.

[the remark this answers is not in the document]
anthonyburdi: Simplified to use defaultdict and added a test case.

great_expectations_cloud/agent/event_handler.py
```diff
@@ -29,11 +31,49 @@
...
+        f"Registered event action: {event_type_str} -> {action_class.__name__} (version {version})"
+    )
+
+
+register_event_action("0", RunCheckpointEvent, RunCheckpointAction)
+register_event_action("0", DraftDatasourceConfigEvent, DraftDatasourceConfigAction)
+register_event_action("0", RunColumnDescriptiveMetricsEvent, ColumnDescriptiveMetricsAction)
+register_event_action("0", ListTableNamesEvent, ListTableNamesAction)
+register_event_action("0", RunMissingnessDataAssistantEvent, RunMissingnessDataAssistantAction)
+register_event_action("0", RunOnboardingDataAssistantEvent, RunOnboardingDataAssistantAction)
```
anthonyburdi, line 74: Actions were not getting registered when these registration calls were in the action specific files.

great_expectations_cloud/agent/event_handler.py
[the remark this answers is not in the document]
Kilo59: Yeah, this is part of the mypy `strict` mode that I turned on a few weeks ago.

- https://github.com/great-expectations/cloud/pull/132

tests/agent/test_event_handler.py
```diff
@@ -20,76 +46,152 @@
...
+        mock_context,
+        mocker: MockerFixture,
+        event_name: str,
+        event: Event,
+        action_type: type[AgentAction[Any]],
+    ):
+        action = mocker.MagicMock(autospec=action_type)
+        mocker.patch("great_expectations_cloud.agent.event_handler._GX_MAJOR_VERSION", "1")
+
+        mocker.patch.dict(_EVENT_ACTION_MAP, {"1": {event_name: action}})
```
Shinnnyshinshin, line 114: From synchronous review: this change triggered the parameterization of the test.

tests/agent/test_event_handler.py
```diff
@@ -20,76 +46,152 @@
...
+
+    def test__get_major_version_raises_on_invalid_version(self):
+        with pytest.raises(InvalidVersionError):
+            with warnings.catch_warnings():
+                # Filter Deprecation warnings about LegacyVersion
+                warnings.filterwarnings(
+                    "ignore",
+                    message="Creating a LegacyVersion has been deprecated and will be removed in the next major release",
+                )
+                _get_major_version("invalid_version")
```
Shinnnyshinshin, line 197: Thank you for drilling down here, and only checking for the `InvalidVersionError` by filtering the warning :bow:

Shinnnyshinshin (approved):

great_expectations_cloud/agent/event_handler.py
```diff
@@ -75,3 +106,29 @@ def handle_event(self, event: Event, id: str) -> ActionResult:
...
+
+class NoVersionImplementationError(Exception):
+    ...
+
+
+class InvalidVersionError(Exception):
+    ...
+
+
+def _get_major_version(version: str) -> str:
```
Kilo59, line 119:
```suggestion
def _get_major_version(version: str | Version) -> str:
```

great_expectations_cloud/agent/event_handler.py
Kilo59: Move this to an `__init__` file, and these will be called.

great_expectations_cloud/agent/event_handler.py
```diff
@@ -29,11 +31,49 @@
...
+#     "1": {
+#         # No events implemented yet for GX Core v1
+#         # Suggested naming convention, add version suffix "V1" e.g.: RunCheckpointEventV1
+#     },
+# }
+_EVENT_ACTION_MAP: dict[str, dict[str, type[AgentAction[Any]]]] = defaultdict(dict)
+
+
+def register_event_action(
+    version: str, event_type: type[Event], action_class: type[AgentAction[Any]]
```
Kilo59, line 55:
```suggestion
    version: str | Version, event_type: type[Event], action_class: type[AgentAction[Any]]
```

Kilo59 (approved):
Couple minor suggestions you can take or leave.
One thing I would like to see before merge is a common `GxAgentEventError` base used for all the custom exceptions.

rreinoldsc (approved):

tests/agent/test_event_handler.py
```diff
@@ -20,76 +46,152 @@
...
+        ],
     )
-    correlation_id = "74842258-803a-48ca-8921-eaf2802c14e2"
-    handler = EventHandler(context=mock_context)
+    def test_register_event_action(self, version: str):
+        register_event_action(version, DummyEvent, DummyAction)
+        assert _EVENT_ACTION_MAP[version][DummyEvent.__name__] == DummyAction
 
-    handler.handle_event(event=event, id=correlation_id)
+        del _EVENT_ACTION_MAP[version][DummyEvent.__name__]  # Cleanup
```
Kilo59, line 159: This is a dangerous way to undo this patch because it won't be hit if the assert above it is triggered.
I would suggest using `pytest` `monkeypatch.setitem()`

https://docs.pytest.org/en/latest/how-to/monkeypatch.html#monkeypatching-dictionaries

anthonyburdi: What about `mocker.patch.dict` instead? Just to keep it consistent by using `mocker` https://pytest-mock.readthedocs.io/en/latest/usage.html

anthonyburdi: Fixed - good call on the suggestion.

great_expectations_cloud/agent/event_handler.py
anthonyburdi: At this point the version is a string, it's parsed in this function.

great_expectations_cloud/agent/actions/__init__.py
```diff
@@ -5,10 +5,15 @@
...
     CreatedResource,
 )
+
+# Import all actions to register them:
 from great_expectations_cloud.agent.actions.data_assistants import (
     RunMissingnessDataAssistantAction,
     RunOnboardingDataAssistantAction,
 )
+from great_expectations_cloud.agent.actions.draft_datasource_config_action import (
+    DraftDatasourceConfigAction,
```
anthonyburdi, line 15: This action was missing from the imports.

PR #151, opened 2024-02-29
ZEL-535: Fix CDM error - pass metrics retriever as a list to avoid iterable error

JennyTee:
Issue:

Running CDM was returning `'ColumnDescriptiveMetricsMetricRetriever' object is not iterable` error after changes in #149 

Solution:

The BatchInspector constructor expects a list of `MetricRetriever` - pass in a list instead

PR #152, opened 2024-02-29
Adds unit test for CDM BatchInspector bug

rreinoldsc:
Added test. Without fix, fails. With fix passes

Follow-up to bugfix in https://github.com/great-expectations/cloud/pull/151


JennyTee (approved):
I <3 this - thank you so much for thinking of and adding!

PR #154, opened 2024-03-04
Unknown Event triggers AMQP Error Response

rreinoldsc:


tests/agent/message_service/test_subscriber.py
```diff
@@ -62,3 +71,30 @@ def test_subscriber_close_handles_amqp_errors_from_connection(error):
     subscriber = Subscriber(client=client)
 
     subscriber.close()  # no exception
+
+def test_subscriber_parse_event_extra_field(mock_subscriber, example_event):
+    event_dict = dict(example_event)
+    event_dict["new_field"] = "surprise!"
+    serialized_bytes = json.dumps(dict(event_dict), indent=2).encode('utf-8')
```
rreinoldsc: is there a better way to do this?

tests/agent/test_event_handler.py
```diff
@@ -18,20 +31,46 @@
...
+    assert isinstance(action, UnknownEventAction)
 
 
 def test_event_handler_handles_run_missingness_data_assistant_event(mocker):
-    action = mocker.patch(
-        "great_expectations_cloud.agent.event_handler.RunMissingnessDataAssistantAction"
-    )
+    mock_action = MagicMock(autospec=RunMissingnessDataAssistantAction)
+    # Override with mock
+    EventHandler._EVENT_TO_ACTION_MAP[RunMissingnessDataAssistantEvent] = mock_action
```
Kilo59, line 72: You don't need to use `MagicMock` for this.
You can & should use the `mocker` fixture because it automatically remove the patches at the end of the test.

https://pytest-mock.readthedocs.io/en/latest/usage.html
```python
def test_event_handler_handles_run_missingness_data_assistant_event(mocker):
    action = mocker.patch(
        "great_expectations_cloud.agent.event_handler.RunMissingnessDataAssistantAction", autospec=RunMissingnessDataAssistantAction
    )

```

great_expectations_cloud/agent/actions/unknown.py
```diff
@@ -0,0 +1,22 @@
...
+)
+from great_expectations_cloud.agent.models import (
+    UnknownEvent,
+)
+
+
+class UnknownEventAction(AgentAction[UnknownEvent]):
+    @override
+    def run(self, event: UnknownEvent, id: str) -> ActionResult:
+        # noop
```
Kilo59, line 17: I think we should emit a warning here.
https://docs.python.org/3/library/warnings.html

great_expectations_cloud/agent/event_handler.py
```diff
@@ -39,31 +40,26 @@ class EventHandler:
...
     """
 
+    _EVENT_TO_ACTION_MAP: Final[dict[type[Event], type[AgentAction[Any]]]] = {
+        RunOnboardingDataAssistantEvent: RunOnboardingDataAssistantAction,
+        RunMissingnessDataAssistantEvent: RunMissingnessDataAssistantAction,
+        ListTableNamesEvent: ListTableNamesAction,
+        RunCheckpointEvent: RunCheckpointAction,
+        RunColumnDescriptiveMetricsEvent: ColumnDescriptiveMetricsAction,
+        DraftDatasourceConfigEvent: DraftDatasourceConfigAction,
+    }
```
Kilo59, line 50:
```suggestion
    _EVENT_TO_ACTION_MAP: ClassVar[dict[type[Event], type[AgentAction[Any]]]] = {
        RunOnboardingDataAssistantEvent: RunOnboardingDataAssistantAction,
        RunMissingnessDataAssistantEvent: RunMissingnessDataAssistantAction,
        ListTableNamesEvent: ListTableNamesAction,
        RunCheckpointEvent: RunCheckpointAction,
        RunColumnDescriptiveMetricsEvent: ColumnDescriptiveMetricsAction,
        DraftDatasourceConfigEvent: DraftDatasourceConfigAction,
    }
```

great_expectations_cloud/agent/models.py
```diff
@@ -10,7 +10,8 @@
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
     class Config:
-        extra: str = Extra.ignore
+        # 2024-03-04: ZEL-501 Strictly enforce models for handling outdated APIs
+        extra: str = Extra.forbid
```
Kilo59, line 14: Have you talked with @JennyTee about this change?
I'm all for it but want to make sure she's in the loop.

Kilo59 (changes requested):
Please use the `pytest-mock` `mocker` fixture instead of using `Mock/MagicMock` directly.
It ensures that mocks are removed after each test.
https://pytest-mock.readthedocs.io/en/latest/index.html

great_expectations_cloud/agent/actions/unknown.py
anthonyburdi: Agree

great_expectations_cloud/agent/models.py
anthonyburdi: :point_up:

great_expectations_cloud/agent/agent.py
```diff
@@ -230,17 +228,29 @@ def _handle_event_as_thread_exit(
...
-                success=True,
-                created_resources=result.created_resources,
-            )
-            print(f"Completed job: {event_context.event.type} ({event_context.correlation_id})")
+
+            if result.type == UnknownEvent().type:
+                status = JobCompleted(
+                    success=False,
+                    created_resources=[],
+                    error_stack_trace="Unknown event. Ensure agent is up-to-date.",
```
JennyTee, line 236: Anna left a comment on ZEL-501 suggesting that the error message should read `The version of the GX Agent you are using does not support this functionality. Please upgrade to {version number} or later.` I think we should run the wording past her--might need to tweak her suggested message if we can't get the version number they should update to in there.

great_expectations_cloud/agent/message_service/subscriber.py
```diff
@@ -147,6 +143,14 @@ def _on_message_handler(
...
         return on_message(event_context)
 
+    @classmethod
+    def parse_event_from(cls, msg_body: bytes) -> Event:
+        try:
+            event: Event = pydantic.parse_raw_as(Event, msg_body)
+            return event
+        except (pydantic.ValidationError, JSONDecodeError):
+            return UnknownEvent()
+
```
anthonyburdi, line 153: Ooh I like this as a separate method. Any reason why this is a `classmethod` on the Subscriber? It's kind of a builder method for an Event, I'm thinking it would be better as a `classmethod` on the Event class or module level method in the file where Event is defined. This would also help with separation of concerns (Subscriber: "Manage an open connection to an event stream." vs an event.)

anthonyburdi (changes requested):
Looking good, just a few things to address

tests/agent/test_event_handler.py
rreinoldsc: I do now see that overriding with the mock affects subsequent tests. However, mocker.patch this did not work because class attribute `EventHandler._EVENT_TO_ACTION_MAP` gets loaded before the mock is set. 

When I use the patch you provided, I get the real object and test fails.

Check out this use of mocker against the map: b422df3f15a937ccc91107f79fbb7c5270fd64d7

great_expectations_cloud/agent/message_service/subscriber.py
rreinoldsc: It was on Subscriber from inertia. I moved it to EventHandler as class method in 6da1c56

great_expectations_cloud/agent/agent.py
rreinoldsc: Set it to closer msg in a0715a9332d0ef27f2e44066119f23f9a281c1fc, confirming with Anna now

great_expectations_cloud/agent/models.py
JennyTee: Thanks, @Kilo59, much appreciated! Glad that we are changing it back now that we can more gracefully handle mismatched event models. 🎉

great_expectations_cloud/agent/actions/unknown.py
rreinoldsc: added warning here, then an error upon failed parse in 350b8b5

Kilo59: This is not a real warning.
`LOGGER.warning` `!=` `warnings.warn()`

rreinoldsc: oh oversight. looking at it now

tests/agent/test_event_handler.py
Kilo59: Looking at that commit it looks like the tests passed.
`EventHandler._EVENT_TO_ACTION_MAP` being defined before the mock shouldn't matter.
Can you link to the failure?

great_expectations_cloud/agent/actions/unknown.py
Kilo59: You'll likely also need to catch this warning in tests.
I would also suggest creating a custom `class GxAgentWarning(UserWarning):...`
https://docs.pytest.org/en/4.6.x/warnings.html#warns
```python
    with pytest.warns(GxAgentWarning):
        code_that_will_throw_this_warning()
```

great_expectations_cloud/agent/event_handler.py
```diff
@@ -72,6 +72,16 @@ def handle_event(self, event: Event, id: str) -> ActionResult:
...
         return action_result
 
+    @classmethod
+    def parse_event_from(cls, msg_body: bytes) -> Event:
+        try:
+            event: Event = pydantic.parse_raw_as(Event, msg_body)
+            return event
+        except (pydantic.ValidationError, JSONDecodeError):
+            # Log as bytes
+            LOGGER.error("Unable to parse event type", extra={"msg_body": f"{msg_body!r}"})
```
Kilo59, line 82:
```suggestion
            LOGGER.warning("Unable to parse event type", extra={"msg_body": f"{msg_body!r}"})
```
We might want this one to be a `warning` level log.
This is expected and doesn't necessarily indicate an unrecoverable error.
https://stackoverflow.com/a/2031209/6304433

tests/agent/test_event_handler.py
Kilo59: I've created a `PR` to provide some guidance around this with the `ruff` `banned-api` rule.
- https://github.com/great-expectations/cloud/pull/158

Kilo59: If for some reason this still doesn't work we can do an inline `noqa: TID251 # comment with explanation` ignore

anthonyburdi: FYI I did a refactor of most of these tests here in case it's helpful: https://github.com/great-expectations/cloud/pull/150/files#diff-73699a97bab49c907e00515541bf18c84ff213b90545dd6f139fcdafd3ecc7b3

great_expectations_cloud/agent/agent.py
rreinoldsc: approved

great_expectations_cloud/agent/actions/unknown.py
rreinoldsc: Check out [new user warning type](https://github.com/great-expectations/cloud/pull/154/files#diff-a89ee08302d15cae4b00da96e8840a01a643d2cb0a82d83aa606a1b300457c9eR6), [raise warning](https://github.com/great-expectations/cloud/pull/154/files#diff-a89ee08302d15cae4b00da96e8840a01a643d2cb0a82d83aa606a1b300457c9eR10), and  [test](https://github.com/great-expectations/cloud/pull/154/files#diff-654cf3c89d7235a8c47412339c29ae838d0e8fd9e78de7e08bc6e10b89c0ba83R12)

rreinoldsc:
blocking on merge of https://github.com/great-expectations/cloud/pull/150

tests/agent/test_event_handler.py
rreinoldsc: fyi going to pull in Anthony's tests which dont have this testing issue. Once merged into main, ill pull into this.

Kilo59:
I'm still seeing CI errors and merge conflicts

rreinoldsc:
ready for re-review @Kilo59 @anthonyburdi

great_expectations_cloud/agent/agent.py
```diff
@@ -176,9 +176,7 @@ def _handle_event_as_thread_enter(self, event_context: EventContext) -> None:
             # this event has been redelivered too many times - remove it from circulation
             event_context.processed_with_failures()
             return
-        elif self._can_accept_new_task() is not True or isinstance(
-            event_context.event, UnknownEvent
-        ):
+        elif self._can_accept_new_task() is not True:
             # request that this message is redelivered later. If the event is UnknownEvent
             # we don't understand it, so requeue it in the hope that someone else does.
```
anthonyburdi, line 181: Minor nit: Does this comment still apply? If not let's remove it or update it.

great_expectations_cloud/agent/agent.py
```diff
@@ -231,17 +229,29 @@ def _handle_event_as_thread_exit(
...
-                success=True,
-                created_resources=result.created_resources,
-            )
-            print(f"Completed job: {event_context.event.type} ({event_context.correlation_id})")
+
+            if result.type == UnknownEvent().type:
+                status = JobCompleted(
+                    success=False,
+                    created_resources=[],
+                    error_stack_trace="The version of the GX Agent you are using does not support this functionality. Please upgrade to latest.",
```
anthonyburdi, line 237: Non blocking suggestion: add an error code here e.g. `error_code="agent-unknown-event"`

tests/agent/test_event_handler.py
```diff
@@ -40,21 +39,32 @@
...
 class TestEventHandler:
-    def test_event_handler_raises_for_unknown_event(self, mock_context):
+    def test_event_handler_unknown_event(self, mocker):
         event = UnknownEvent()
         correlation_id = "74842258-803a-48ca-8921-eaf2802c14e2"
-        handler = EventHandler(context=mock_context)
-
-        with pytest.raises(UnknownEventError):
-            handler.handle_event(event=event, id=correlation_id)
+        context = mocker.MagicMock(autospec=CloudDataContext)
```
anthonyburdi, line 63: Not a blocker, but you can alternatively use the `mock_context` fixture in this test.

anthonyburdi (approved):
LGTM! Just a few non blocking suggestions.

great_expectations_cloud/agent/agent.py
rreinoldsc: updated in 477073b

great_expectations_cloud/agent/agent.py
rreinoldsc: id like to move to another PR

tests/agent/test_event_handler.py
rreinoldsc: updated in 04c5f68

rreinoldsc:
Addressed Gabriels feedback

tests/agent/test_event_handler.py
```diff
@@ -193,3 +201,37 @@ def test__get_major_version_raises_on_invalid_version(self):
...
+    event_dict["new_field"] = "surprise!"
+    serialized_bytes = json.dumps(dict(event_dict), indent=2).encode("utf-8")
+    event = EventHandler.parse_event_from(serialized_bytes)
+
+    assert event.type == "unknown_event"
+
+
+def test_parse_event_missing_required_field(example_event):
+    event_dict = example_event.dict()
+    del event_dict["datasource_name"]
```
Kilo59, line 217:
```suggestion
    event_dict = example_event.dict(exclude={"datasource_name"})
```

rreinoldsc: oh this is smooth

rreinoldsc: impl [here](https://github.com/great-expectations/cloud/pull/164/commits/481bbf0d4dbe4163aff47702a5a2b68cfcadeed6) in this PR: https://github.com/great-expectations/cloud/pull/164/commits/481bbf0d4dbe4163aff47702a5a2b68cfcadeed6

PR #158, opened 2024-03-04
Ban `unittest.mock` for tests

Kilo59:
## Ban `unittest.mock`

Without using the `with unittest.mock.patch()` context manager, unittest patches/mocks do not clean themselves up at the end of tests.
This is dangerous as tests can end up having more mocks applied than is intended.

We can catch these issues by using `TID251` `banned-ap` rules to prevent the use of `unittest.mock` `Mock`, `MagicMock` & `patch` and instead suggest using the `pytest_mock` `mocker` fixture or `pytest monkeypatch` which both automatically cleanup after themselves.

https://pytest-mock.readthedocs.io/en/latest/index.html
https://docs.pytest.org/en/latest/how-to/monkeypatch.html


tests/ruff.toml
```diff
@@ -26,3 +24,10 @@ lint.extend-ignore = [
 
 [lint.isort]
 known-first-party = ["great_expectations_cloud", "tests"]
+
+[lint.flake8-tidy-imports.banned-api]
+# https://pytest-mock.readthedocs.io/en/latest/index.html
+# standard library mock doesn't clean up after itself which can affect unrelated tests
+"unittest.mock.MagicMock".msg = "Please do not use unittest.mock.MagicMock directly, use the `pytest-mock` `mocker` fixture or `pytest.monkeypatch` instead."
+"unittest.mock.Mock".msg = "Please do not use unittest.mock.Mock directly, use the `pytest-mock` `mocker` fixture or `pytest.monkeypatch` instead."
+"unittest.mock.patch".msg = "Please do not use unittest.mock.patch directly, use the `pytest-mock` `mocker` fixture or `pytest.monkeypatch` instead."
```
Kilo59, line 33: Banning `unittest.mock` for tests

tests/ruff.toml
```diff
@@ -8,8 +8,6 @@ lint.extend-ignore = [
     # This is likely to be a high-touch rule. Doing this in `tests` doesn't help circular imports.
     # Let's differ this for tests until there are auto-fixes
     "TCH001",
-    # https://beta.ruff.rs/docs/rules/#flake8-tidy-imports-tid
-    "TID251",  # banned-api - is used to prevent opaque attribute errors caused by missing optional dependencies
```
anthonyburdi, line 12: Is the existing functionality of TID251 still enabled with the addition of the `[lint.flake8-tidy-imports.banned-api]` section?

tests/agent/conftest.py
```diff
@@ -37,6 +41,12 @@ def _mock_get_lastest_version_from_pypi(self) -> Version:
     )
 
 
+@pytest.fixture
+def mock_context(mocker: MockerFixture) -> CloudDataContext:
```
anthonyburdi, line 45: 🚀

anthonyburdi (commented):
LGTM - just one question

tests/ruff.toml
anthonyburdi: Nevermind - realizing now you are removing from `lint.extend-ignore` i.e. you are enabling.

anthonyburdi (approved):
Thank you for enabling this rule!

tests/ruff.toml
Kilo59: No, I'm overriding the `[lint.flake8-tidy-imports.banned-api]` rules.
To extend those rules I'd have to add them to the overridden section.
We shouldn't need to care about the compatibility imports in tests.

PR #163, opened 2024-03-06
Add instructions for adding an action

anthonyburdi:
Add instructions to the readme for adding an action to the agent.


JennyTee (approved):
lgtm! I made a quick update to the numbering. thanks for adding this, good call.

anthonyburdi:
> lgtm! I made a quick update to the numbering. thanks for adding this, good call.

Thank you! FYI I did the numbering as 1. 1. 1. so that if we add more items then the numbers are updated automatically. But it's only three items so not a big deal either way, merging!

PR #180, opened 2024-03-20
Force version bump on poetry.lock change

anthonyburdi:
Make sure we bump the agent version via CI check when bumping dependencies e.g. GX version in `poetry.lock`.


Shinnnyshinshin (approved):
Great suggestion @anthonyburdi

PR #189, opened 2024-04-02
catch probable auth error and enables retry

allensallinger:
catch probable auth error and enables retry
test for probable auth error has been added as well.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -123,7 +123,7 @@ class GXAgent:
     #          by adding a delay and retrying the connection. Retrying with new credentials
     #          requires calling get_config again, which handles the password generation.
     @retry(
-        retry=retry_if_exception_type(AuthenticationError),
+        retry=retry_if_exception_type((AuthenticationError, ProbableAuthenticationError)),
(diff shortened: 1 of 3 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -144,7 +144,7 @@ def _listen(self) -> None:
             print("Received request to shutdown.")
         except (SubscriberError, ClientError):
             print("Connection to GX Cloud has encountered an error.")
-        except AuthenticationError:
+        except (AuthenticationError, ProbableAuthenticationError):
```
anthonyburdi, line 147: I think we need to also add `ProbableAuthenticationError` to the `@retry` decorator above on line 126 to enable the retry: https://tenacity.readthedocs.io/en/stable/api.html#tenacity.retry.retry_if_exception_type

anthonyburdi (changes requested):
Thanks for handling! Looks good just one small edit.

great_expectations_cloud/agent/agent.py
allensallinger: Ahh, fixed that and fixed up the tests. Thanks for the review.

anthonyburdi (approved):
Thanks for fixing!

PR #193, opened 2024-04-09
use logger with minimized cardinality

rreinoldsc:
Moves subset of #print usage to logger


anthonyburdi (commented):
LGTM. Just commenting since it's in draft. Also can we wait to merge until after demos today?

anthonyburdi:
> LGTM. Just commenting since it's in draft. Also can we wait to merge until after demos today?

Also what do you think about making this a pre-release? If you think the change can't wait until next thurs then it definitely can be a release, but we should default to pre-releases for other changes.

rreinoldsc:
@anthonyburdi  I'd like to get it running in our internal agents. If those are on pre-release, then I can make it a pre-release. Otherwise, full release

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "20240424.0"
+version = "20240425.0"
```
anthonyburdi, line 3:
```suggestion
version = "20240426.0"
```

anthonyburdi (approved):
Just need to bump version again but otherwise LGTM.

pyproject.toml
rreinoldsc: i upped it

PR #197, opened 2024-04-12
Invoke script updates for the new versioning scheme

anthonyburdi:
Update the release logic to match the new release process documented in https://github.com/great-expectations/cloud/pull/195

Example usage:

`invoke release`:

image

`invoke pre-release`:

image


tasks.py
```diff
@@ -74,7 +76,7 @@ def lint(ctx: Context, check: bool = False, unsafe_fixes: bool = False) -> None:
 
 
 @invoke.task(
-    aliases=["build"],
+    aliases=("build",),
```
anthonyburdi, line 79: `invoke` is looking for a tuple here instead of a list. Noticed in the new code, fixed all the existing usages in this file.

tasks.py
```diff
@@ -154,24 +156,111 @@ def _get_latest_version() -> Version:
...
+    proposed_version = Version(current_date)
+    if proposed_version > latest_version:
+        # Always bump if it is a new date
+        return Version(f"{proposed_version.major}.0")
+    else:
+        # Bump the minor version only if it is the same date
+        return Version(f"{latest_version.major}.{latest_version.minor + 1}")
+
+
+class PreReleaseVersionError(Exception):
```
anthonyburdi, line 186: This custom exception might be overkill, happy to remove.

tasks.py
```diff
@@ -154,24 +156,111 @@ def _get_latest_version() -> Version:
...
+    elif latest_pre_release_version.major < int(current_date):
+        new_version = Version(f"{current_date}.0.dev0")
+        return new_version
+    else:
+        # We should never get here
+        raise PreReleaseVersionError(
+            latest_version=latest_version,
+            latest_pre_release_version=latest_pre_release_version,
+            current_date=current_date,
+        )
```
anthonyburdi, line 229: I simplified this logic quite a bit, but happy to simplify further if there is any low hanging fruit.

tests/test_project.py
```diff
@@ -78,26 +82,202 @@ def test_pre_commit_versions_are_in_sync(
...
+        ).params(),
+        BumpVersionParams(
+            id="pre-release from pre-release semver 0.0.1.dev1 -> 20240411.0.dev0",
+            expected_version=Version("20240411.0.dev0"),
+            pre_release=True,
+            latest_version=Version("0.0.1"),
+            latest_pre_release_version=Version("0.0.1.dev1"),
+            current_date="20240411",
+        ).params(),
     ],
```
anthonyburdi, line 256: Did I miss any test cases in this list?

tasks.py
```diff
@@ -196,21 +285,31 @@ def _update_version(version_: Version | str) -> None:
...
 
-    print("\n  bumping version ...", end=" ")
-    new_version = bump_version(local_version, pre_release=pre)
+    print(f"\nbumping version ({release_type}) ...", end=" ")
+
+    new_version = bump_version(
+        pre_release=pre,
+        latest_version=latest_release_version,
+        latest_pre_release_version=latest_pre_release_version,
+        current_date=datetime.date.today().strftime("%Y%m%d"),  # noqa: DTZ011 # timezone agnostic, local time OK.
```
Kilo59, line 306: I think we should use timezone aware UTC time here. It totally eliminates an entire class of problems relating to multiple releases in a short timespan that go back and forth between a dateline.

tasks.py
```diff
@@ -154,24 +156,111 @@ def _get_latest_version() -> Version:
...
-            new_version = Version(f"{version_.major}.{version_.minor}.{version_.micro + 1}")
-    elif version_.dev:
-        # bump an existing pre-release version
-        new_version = Version(
-            f"{version_.major}.{version_.minor}.{version_.micro}.dev{version_.dev + 1}"
+@functools.lru_cache(maxsize=1)
+def _get_all_versions() -> list[Version]:
+    r = requests.get("https://pypi.org/pypi/great-expectations-cloud/json", timeout=10)
+    r.raise_for_status()
+    return [Version(v) for v in r.json()["releases"].keys()]
```
Kilo59, line 163:
```suggestion
    return [Version(v) for v in r.json()["releases"]]
```

tasks.py
Kilo59: My only suggestion would be spending a bit of time looking into existing python packages that can do this thing. It feels error prone. We might still decide this is fine as it is.
Like what does `Versions.__add__` do?

Or something like `bump-my-version` (formerly `bump-version`) should work, but you've already done a bunch of the work.
https://callowayproject.github.io/bump-my-version/

anthonyburdi: I'll also review your comments more thoroughly but I did look into [bumpver](https://github.com/mbarkhau/bumpver) and a few others that turned out to be not very well maintained. It seemed like it would be a little complicated to do what we are planning here in bumpver so I just implemented. I can investigate bump-my-version, but figure this is pretty simple and worst case we can bump manually if any of the logic is off when you run the tool so figure it is relatively low priority to migrate.

tasks.py
anthonyburdi: Good point - I was mainly worried about someone releasing later in the day and seeing the date be the next day because it is "tomorrow" in UTC. But you're right - I agree that standardizing is better so I'll make this change.

anthonyburdi: To handle the "tomorrow" issue ☝️ I was going to set to ET but didn't want to add another dependency just for that.

tasks.py
anthonyburdi: Thanks for the input here @TrangPham @Kilo59 - I simplified the logic to only use the latest release. Let me know what you think!
I think for now if you are good with the logic I'd like to release this just to keep moving forward and investigate more deeply into replacing with bump-my-version after we get a bit more of the epic work done.

anthonyburdi: (Mainly because I'm not seeing an obvious way right off the bat to incorporate pre-releases with calver in bump-my-version: https://callowayproject.github.io/bump-my-version/howtos/calver/)

Kilo59 (approved):

PR #203, opened 2024-04-16
[MAINTENANCE] Update ci to check version is not bumped and then auto bump pre-release on merge

TrangPham:
How this works:

Agent CI and Merge Flow Chart


Kilo59 (commented):
What's wrong with the diff?

.github/workflows/ci.yaml
Kilo59, line 1: Can we only trigger a need for release if a core dependency changes (or at least ignore dev Dependency changes).
Like we don't want a new pre-release if the version of pytest changes.
Ideally we would also ignore things like formatting changes or like minor annotation changes but that would involve things like comparing asts (which seems like way too much effort).

.github/workflows/ci.yaml
```diff
@@ -6,6 +6,81 @@ on:
...
+        task: ["lint", "fmt", "type-check", "python-build", "docker"]
+      fail-fast: false
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
```
Kilo59, line 21: What's going on with the diff here?

Kilo59 (commented):
This looks great.
Minor question about the diff and question about ignoring dev-dependency changes.

Kilo59:
@TrangPham, what did you use to create the diagram?

.github/workflows/containerize-agent.yaml
```diff
@@ -103,8 +103,24 @@ jobs:
         id: get_version
         run: echo "POETRY_VERSION=$(poetry version -s)" >> $GITHUB_OUTPUT
 
+      - name: Get is_release from pyproject.toml
+        id: get_is_release
+        run: echo "IS_RELEASE=$(poetry version -s | grep -q 'dev' && echo false || echo true)" >> $GITHUB_OUTPUT
```
anthonyburdi, line 108: Nice!

.github/workflows/containerize-agent.yaml
```diff
@@ -103,8 +103,24 @@ jobs:
...
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: |
+            greatexpectations/agent:${{ steps.get_version.outputs.POETRY_VERSION }}
+            258143015559.dkr.ecr.us-east-1.amazonaws.com/gx/agent:${{ steps.get_version.outputs.POETRY_VERSION }}
+
+      - name: Push Pre-Release to Docker Hub and Internal ECR
+        if: ${{ steps.get_is_release.outputs.IS_RELEASE == 'false' }}
```
anthonyburdi, line 123: Should this section be for release not pre-release?
```suggestion
      - name: Push Release to Docker Hub and Internal ECR
        if: ${{ steps.get_is_release.outputs.IS_RELEASE == 'true' }}
```

anthonyburdi (changes requested):
Looks good, just one minor suggestion

.github/workflows/ci.yaml
anthonyburdi: Sounds like a good plan. Let's not block this PR on that though unless @TrangPham already has a solution cooking :)

anthonyburdi (approved):

.github/workflows/ci.yaml
anthonyburdi: @Kilo59 - for future reference would a potential solution be something like parsing the pyproject.toml dependency sections individually and looking for changes only to core dependency changes for deciding whether to bump?

PR #204, opened 2024-04-16
Z2-296: Add stable tag to dockerhub

JennyTee:
Add `stable` tag to agent image that is pushed to dockerhub.

(Note: we are keeping the `latest` tag for now.)

### DIFF
--- a/.github/workflows/containerize-agent.yaml
+++ b/.github/workflows/containerize-agent.yaml
@@ -64,7 +64,9 @@ jobs:
           platforms: linux/arm64
           push: false
           load: false
-          tags: greatexpectations/agent:latest
+          tags: |
+            greatexpectations/agent:latest
+            greatexpectations/agent:stable
 
       # https://docs.docker.com/build/ci/github-actions/test-before-push/
       - name: Build amd64, With Load, No Push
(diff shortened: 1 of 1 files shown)


.github/workflows/containerize-agent.yaml
JennyTee, line 1: Is there a way to test these changes?

anthonyburdi: Since it is additive I think we can test on the next release since it will not break any existing deploys. I'm not sure if there is a better way. @rreinoldsc do you have any ideas on how we could potentially test this without an actual release?

anthonyburdi (approved):
LGTM!

.github/workflows/containerize-agent.yaml
```diff
@@ -129,6 +133,7 @@ jobs:
           tags: |
             greatexpectations/agent:${{ steps.get_version.outputs.POETRY_VERSION }}
             greatexpectations/agent:latest
+            greatexpectations/agent:stable
```
allensallinger, line 136: Is there ever going to be a time when the `latest` and `stable` images are not the same?

JennyTee: No, not to my knowledge--with these changes, `stable` and `latest` would always be the same.

JennyTee: (The thinking here was to keep `latest` just in case users try to pull an image without specifying a tag, since `docker pull` defaults to using the `latest` tag if another tag isn't specified.

allensallinger: That makes sense, is the stable tag present to give peace of mind then to our users?

allensallinger (approved):

.github/workflows/containerize-agent.yaml
JennyTee: Yeah pretty much - we may decide to decouple `stable` and `latest` in the future, but that isn't being considered in this epic. The main change in this epic is that we'll start using pre-release versioning and will only push release version images to dockerhub once/week. (Add'l info here: https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1019838486/Agent+Versioning+Release+Strategy)

PR #205, opened 2024-04-17
Upgrade Docker image to Python `3.10.13`

Shinnnyshinshin:
- Update python version `3.10.12` -> `3.10.13` 
- [Proposed change from 2024-04-17](https://greatexpectations.atlassian.net/wiki/x/B4CcP) to address vulnerabilities.

PR #208, opened 2024-04-17
Update Poetry Lock file after python upgrade in Docker

Shinnnyshinshin:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "20240417.0"
```
Kilo59, line 3: This update makes a lot of dependency changes.
- I would suggest waiting until #204 merges, making this a pre-release, and doing some manual testing before merging this.

Shinnnyshinshin: ok great idea @Kilo59. I'll follow up here tomorrow.

anthonyburdi: Agree. @Kilo59 Would it make sense to revert the docker image bump PR as well in the meantime? I'm thinking yes. https://github.com/great-expectations/cloud/pull/205 Then incorporate that change with this in a pre-release as you suggested.

Dockerfile
```diff
@@ -20,6 +20,7 @@ ENV POETRY_CACHE_DIR=/tmp/pypoetry
 
 # Required for arm64, for building psutil
 RUN apt-get update && apt-get install --no-install-recommends gcc=4:12.2.0-3 -y && rm -rf /var/lib/apt/lists/*
+RUN apt-get apt-get install --no-install-recommends libpq-dev python-dev build-essential libsnappy-dev
```
Kilo59, line 23:
```suggestion
RUN apt-get install --no-install-recommends libpq-dev python-dev build-essential libsnappy-dev
```

Dockerfile
```diff
@@ -20,6 +20,7 @@ ENV POETRY_CACHE_DIR=/tmp/pypoetry
 
 # Required for arm64, for building psutil
 RUN apt-get update && apt-get install --no-install-recommends gcc=4:12.2.0-3 -y && rm -rf /var/lib/apt/lists/*
+RUN apt-get install --no-install-recommends libpq-dev python-dev build-essential libsnappy-dev
 
 RUN pip --no-cache-dir install poetry==1.6.1
```
Kilo59, line 25:
```suggestion
RUN pip --no-cache-dir install poetry==1.8.2
```
Not sure why I didn't notice this earlier, but this is an old version of poetry.

Try reverting the extra `apt-get` commands and seeing if this by itself fixes it.

Kilo59, line 25:
```suggestion

RUN pip --no-cache-dir install poetry==1.8.2
```
Not sure why I didn't notice this earlier, but this is an old version of poetry.

Try reverting the extra `apt-get` commands and seeing if this by itself fixes it.

Dockerfile
Shinnnyshinshin: :) oops. Fixed this one.

Kilo59 (approved):

poetry.lock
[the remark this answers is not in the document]
Shinnnyshinshin: Thank you @Kilo59. This really helps clear up my understanding :)

PR #213, opened 2024-04-18
[MAINTENANCE] Update with flow diagram

TrangPham:


.github/workflows/agent_release_workflows.png
Kilo59, line 1: I think this is great as is.
But if we wanted to we could do this as a mermaid diagram, and then the diagram would update if and when changes were made.
https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

Kilo59 (approved):
🙌

PR #217, opened 2024-04-19
Maintenance - Retire `ColumnDescriptiveMetricsAction`

Shinnnyshinshin:
- Retiring `ColumnDescriptiveMetricsAction` from Agent code, since we are now using `MetricListAction`.


tests/agent/test_event_handler.py
```diff
@@ -101,11 +104,13 @@ def test_event_handler_unknown_event(self, mock_context):
...
             ),
             (
-                "RunColumnDescriptiveMetricsEvent",
-                RunColumnDescriptiveMetricsEvent(
-                    datasource_name="test-datasource", data_asset_name="test-data-asset"
+                "RunMetricListEvent",
+                RunMetricListEvent(
+                    datasource_name="test-datasource",
+                    data_asset_name="test-data-asset",
+                    metrics=[MetricTypes.TABLE_COLUMN_TYPES, MetricTypes.TABLE_COLUMNS],
```
Shinnnyshinshin, line 111: Updated `event_hander` test to include `RunMetricListEvent`

JennyTee (commented):
Not trying to create extra work, but imo it would also be helpful if we updated `CreateColumnDescriptiveMetricsJob` in the mercury-web code to be `MetricListJob` or similar. Not super high priority; doesn't have any effect on the code, but I was confused for a sec earlier today when looking through that code b/c of the naming difference.

JennyTee (approved):

Shinnnyshinshin:
> Not trying to create extra work, but imo it would also be helpful if we updated `CreateColumnDescriptiveMetricsJob` in the mercury-web code to be `MetricListJob` or similar. Not super high priority; doesn't have any effect on the code, but I was confused for a sec earlier today when looking through that code b/c of the naming difference.

Created [Z2-597](https://greatexpectations.atlassian.net/jira/software/c/projects/Z2/boards/85?selectedIssue=Z2-597) to track this work. I think it's definitely possible to include it as part of the current clean up effort :)

PR #220, opened 2024-04-22
[MAINTENANCE] remove core gx version print out at start of agent

TrangPham:


great_expectations_cloud/agent/agent.py
```diff
@@ -87,9 +87,7 @@ class GXAgent:
 
     def __init__(self: Self):
         agent_version: str = self.get_current_gx_agent_version()
-        great_expectations_version: str = self._get_current_great_expectations_version()
```
JennyTee, line 90: Can we also remove the `_get_current_great_expectations_version()` method? Don't think we will need it for anything else.

JennyTee (approved):

anthonyburdi (commented):
LGTM pending Jenny's request

PR #223, opened 2024-04-23
update readme for clarity re: pyproject.toml updates

JennyTee:
Updating README as discussed in https://github.com/great-expectations/cloud/pull/216#pullrequestreview-2018018976

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,11 +218,11 @@ A visual representation of the workflow is shown [here](./.github/workflows/agen
 ### Dependabot and Releases/Pre-releases
 GitHub's Dependabot regularly checks our dependencies for vulnerabilty-based updates and proposes PRs to update dependency version numbers accordingly.
 
-Dependabot may only update the `poetry.lock` file. Before merging Dependabot suggestions, we should also ensure that `pyproject.toml` is aligned with version locked in the `poetry.lock` file by following the instructions above at [Updating `poetry.lock` dependencies](#updating-poetry.lock-dependencies).
+Dependabot may only update the `poetry.lock` file. If only changes to `poetry.lock` are made, this may be done in a pre-release.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -218,11 +218,11 @@ A visual representation of the workflow is shown [here](./.github/workflows/agen
...
 
-Dependabot may only update the `poetry.lock` file. Before merging Dependabot suggestions, we should also ensure that `pyproject.toml` is aligned with version locked in the `poetry.lock` file by following the instructions above at [Updating `poetry.lock` dependencies](#updating-poetry.lock-dependencies).
-
-Note: if Dependabot suggests an update to a tool in the `[tool.poetry.group.dev.dependencies]` group in `pyproject.toml`, these changes can be merged in a pre-release version (i.e., a standard release is not required). While doing this, make sure any version references in the pre-commit config `.pre-commit-config.yaml` are kept in sync (e.g., ruff).
-
-For other dependency updates, a new release should be orchestrated. This includes updates in the following sections:
-
-- `[tool.poetry.dependencies]`
-- `[tool.poetry.group.*.dependencies]` where `*` is the name of the group (not including the `dev` group)
+Dependabot may only update the `poetry.lock` file. If only changes to `poetry.lock` are made, this may be done in a pre-release.
```
anthonyburdi, line 221: Slight clarification: If changes to the version of `great_expectations` (GX Core) are made in `poetry.lock` then we do need to bump the version. This will be temporary until the work you are doing to remove the `[cloud]` option in GX Core and pin the GX Core version in pyproject.toml is done. Once that is done, we will just change the GX Core version in `pyproject.toml` when a new version of GX Core is released, which will necessitate a release of the agent.

tldr: I think this is not 100% accurate currently but fine for now since we are in the process of removing the cloud dependency from GX Core.

anthonyburdi (approved):
Thank you for updating!

README.md
JennyTee: ah - thanks for the clarification - appreciate the info!

PR #229, opened 2024-04-25
Z2-289: Pin OSS version in pyproject.toml

JennyTee:
* Now that we have removed the optional [cloud] dependency from GX Core, we can pin the Agent to the GX OSS version in pyproject.toml
* The poetry.lock changes in this PR are the result of running `poetry update great-expectations` after making the pyproject update


JennyTee:
closing in favor of https://github.com/great-expectations/cloud/pull/238

PR #237, opened 2024-04-29
Fix project url

rreinoldsc:


Kilo59 (approved):
I think we should have a mechanism to opt-out of a release or version bump.

Maybe the version bump action could check for a `no-release`/`no-bump` label or word in the title. 

Something like a mypy or ruff config change shouldn't trigger a new pypi release for example.

PR #240, opened 2024-04-30
Draft datasource config integration tests

anthonyburdi:
Integration tests to make sure the DraftDatasourceConfigAction runs against a live app instance + GX Core. This is a prerequisite to releasing the GX Agent.

Note: Uses draft datasource configs in seed data added via https://github.com/greatexpectationslabs/mercury/pull/3166

PR #241, opened 2024-04-30
Add integration marker, CI step + codecov flag

anthonyburdi:
Add an integration marker, CI step + codecov flag for tests that rely on a running GX Cloud container.

See request: https://github.com/great-expectations/cloud/pull/240#discussion_r1585517026

Note that this also marks `test_agent.py` as an integration test. We should look into those tests and aim to make them run as unit tests, e.g. by mocking calls to the backend.

Note also that I'm not sure about the codecov flag config here yet so keeping as draft.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,7 +111,89 @@ jobs:
 
       - name: Run tests
 
-        run: poetry run pytest -m "not agentjobs" -W default::great_expectations.datasource.fluent.GxInvalidDatasourceWarning --cov-report xml
+        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml
+
+      # upload coverage report to codecov
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.1.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: great-expectations/cloud
+          flags: ${{ matrix.python-version }}
+
+  integration-tests:
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+    permissions:
+      id-token: write
+      contents: read
+      actions: read
+    runs-on: ubuntu-latest
+
+    env:
+      # GX Agent
+      GX_CLOUD_ACCESS_TOKEN: ${{ secrets.GX_CLOUD_ACCESS_TOKEN }}
+      GX_CLOUD_ORGANIZATION_ID: ${{ secrets.GX_CLOUD_ORGANIZATION_ID }}
+      GX_CLOUD_BASE_URL: "http://localhost:5000"
+      # GX Cloud
+      AUTH0_DOMAIN: ${{secrets.AUTH0_DOMAIN}}
+      AUTH0_API_AUDIENCE: ${{secrets.AUTH0_API_AUDIENCE}}
+      AUTH0_MERCURY_API_CLIENT_ID: ${{secrets.AUTH0_MERCURY_API_CLIENT_ID}}
+      AUTH0_MERCURY_API_CLIENT_SECRET: ${{secrets.AUTH0_MERCURY_API_CLIENT_SECRET}}
+      LOGGING_LEVEL: DEBUG
+      ENVIRONMENT: local
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache-dependency-path: poetry.lock
+
+      - name: Load cached Poetry installation
+        id: cached-poetry
+        uses: actions/cache@v4
+        with:
+          path: ~/.local # the path depends on the OS
+          key: poetry-${{ matrix.python-version  }} # increment to reset cache
+
+      - name: Install Poetry
+        if: steps.cached-poetry.outputs.cache-hit != 'true'
+        uses: snok/install-poetry@v1
+
+      - name: Install dependencies
+        run: poetry install --sync
+
+      - name: Configure ECR AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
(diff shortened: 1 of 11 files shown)


tests/agent/test_agent.py
```diff
@@ -39,6 +39,9 @@
     from tests.agent.conftest import DataContextConfigTD
 
 
+pytestmark = pytest.mark.integration
```
Shinnnyshinshin, line 42: would these qualify as integration tests? I was thinking `unit`

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,47 @@ jobs:
       - run: poetry run invoke ${{ matrix.task }} --check
 
   tests:
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
+    runs-on: ubuntu-latest
```
Shinnnyshinshin, line 46: 🚀

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,47 @@ jobs:
...
+      - name: Run tests
+        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml
+
+      # upload coverage report to codecov
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.1.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: great-expectations/cloud
+          flags: ${{ matrix.python-version }},unit
```
anthonyburdi: I'm not sure I did the codecov flags correctly. Originally I had separate flags e.g. `flags: ${{ matrix.python-version }},unit` which shows as:
image
Then changed to more of a namespace-like approach: `flags: unit-${{ matrix.python-version }}` and `flags: integration-${{ matrix.python-version }}`
Feedback welcome!

tests/agent/test_agent.py
anthonyburdi: I thought so too, but they do require more than the env variables - it looks like they are trying to make a connection: https://github.com/great-expectations/cloud/actions/runs/8903149643/job/24450383609#step:7:726. I agree, after a brief look I think these should be unit tests but if so we probably need to change the tests themselves.

Kilo59: Can you annotate the markers in `init_options` to explain what we are considering an integration test.

`integration: test that relies on a running GX Cloud instance`.

And/or maybe just make the marker name more explicit. `gx_cloud_integration`

.github/workflows/ci.yaml
Kilo59: I don't think we need to run these integration tests on every version of python.
The non-integration tests should catch any python version related problems.

I would just run the integration tests on a single version of python, and leave the python version out of the codecov flag. I would use whatever version we use for the container.

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,47 @@ jobs:
       - run: poetry run invoke ${{ matrix.task }} --check
 
   tests:
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
```
Kilo59, line 45:
```suggestion
        python-version: ["3.11"]
```

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,47 @@ jobs:
...
+      - name: Run tests
+        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml
+
+      # upload coverage report to codecov
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.1.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: great-expectations/cloud
+          flags: unit-${{ matrix.python-version }}
```
Kilo59, line 81:
```suggestion
          flags: ${{ matrix.python-version }}
```
The tests running are more than just unit-tests.
I would leave this out of the flag.

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,47 @@ jobs:
...
+
+  integration-tests:
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
```
Kilo59, line 86:
```suggestion
        python-version: ["3.10"]
```

.github/workflows/ci.yaml
```diff
@@ -111,15 +152,15 @@ jobs:
...
+        run: poetry run pytest -m integration -W default::great_expectations.datasource.fluent.GxInvalidDatasourceWarning --cov-report xml
 
       # upload coverage report to codecov
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4.1.0
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           slug: great-expectations/cloud
-          flags: ${{ matrix.python-version }}
+          flags: integration-${{ matrix.python-version }}
```
Kilo59, line 163:
```suggestion
          flags: integration
```

.github/workflows/ci.yaml
```diff
@@ -111,15 +152,15 @@ jobs:
 
       - name: Run tests
 
-        run: poetry run pytest -m "not agentjobs" -W default::great_expectations.datasource.fluent.GxInvalidDatasourceWarning --cov-report xml
+        run: poetry run pytest -m integration -W default::great_expectations.datasource.fluent.GxInvalidDatasourceWarning --cov-report xml
```
Kilo59, line 155:
```suggestion
        run: poetry run pytest -vv -m integration -W default::great_expectations.datasource.fluent.GxInvalidDatasourceWarning --cov-report xml
```
We shouldn't have that many integration tests so I think using the `verbose` flag can be helpful especially when these tests get slower.

Kilo59 (approved):

Kilo59:
Thanks for this!

tests/agent/test_agent.py
anthonyburdi: Do you happen to know what the `agentjobs` marker is for? It looks like a single test `test_job_processing` which looks like it runs a checkpoint job during the docker image build pipeline. I'm guessing a integration/smoketest?

PR #263, opened 2024-05-17
Update coverage config

Kilo59:
Update coverage settings to reduce `Codecov` CI failures on "good" PRs and make the reporting more useful.

In general slightly reducing the coverage targets needed for a "pass".
Also exclude certain lines for the coverage reporting that we wouldn't expect to be covered in a test such as...

1. `if TYPE_CHECKING:` blocks
2. `raise NotImplementedError`
3. Abstract methods.


codecov.yml
```diff
@@ -1,17 +1,18 @@
 flag_management:
   default_rules: # the rules that will be followed for any flag added, generally
     carryforward: false
+    after_n_builds: 5 # should correspond to the number of coverage reports we expect
```
Kilo59, line 4: Delays the status-check reporting until this number of reports have been uploaded.

This prevents unit tests finishing first and causing `codecov` to complain about missing coverage that will be reported once integration tests finish.

pyproject.toml
```diff
@@ -232,3 +233,17 @@ filterwarnings = [
...
+exclude_also = [
+    # type-checking imports don't exist at runtime
+    "if TYPE_CHECKING:",
+    # Don't complain if tests don't hit NotImplementedError:
+    "raise NotImplementedError",
+    # Don't complain if non-runnable code isn't run:
+    "if __name__ == .__main__.:",
+    # Don't complain about abstract methods, they aren't run:
+    "@(abc\\.)?abstractmethod",
+    ]
```
JennyTee, line 249: nice

JennyTee (approved):
🙏

PR #294, opened 2024-06-06
Create scheduled jobs

anthonyburdi:
This PR sends a request to GX Cloud to create a job using a new endpoint (related: https://github.com/greatexpectationslabs/mercury/pull/3341) and set it started when a scheduled job is handled from the queue. Since scheduled jobs are created by the scheduler + lambda (rather than by GX Cloud) we need to create the job in GX Cloud so that we can use the job to track completion status and show status in the UI.

Also in this PR to enable the above:
- A specific event/action is created for scheduled checkpoints: `RunScheduledCheckpointAction` and `RunScheduledCheckpointEvent`. This helps disambiguate scheduled from non scheduled runs, and adds the `schedule_id` field.
- Logic for running a checkpoint is moved to a module level function so it can be called in both `RunCheckpointAction` and `RunScheduledCheckpointAction`


tests/agent/actions/checkpoints/test_run_checkpoint_action.py
```diff
@@ -19,85 +23,61 @@
...
+            ),
+        ),
+        (
+            RunScheduledCheckpointAction,
+            RunScheduledCheckpointEvent(
+                type="run_scheduled_checkpoint.received",
+                datasource_names_to_asset_names={"Data Source 1": {"Data Asset A", "Data Asset B"}},
+                checkpoint_id="5f3814d6-a2e2-40f9-ba75-87ddf485c3a8",
+                schedule_id=UUID("5f3814d6-a2e2-40f9-ba75-87ddf485c3a8"),
+            ),
```
anthonyburdi, line 55: Avoiding a duplicate set of tests by using parametrization in this test file for both the scheduled and non-scheduled version.

great_expectations_cloud/agent/models.py
```diff
@@ -47,6 +51,13 @@ class RunCheckpointEvent(EventBase):
     splitter_options: Optional[Dict[str, Any]] = None
 
 
+class RunScheduledCheckpointEvent(ScheduledEventBase):
+    type: Literal["run_scheduled_checkpoint.received"] = "run_scheduled_checkpoint.received"
+    datasource_names_to_asset_names: Dict[str, Set[str]]
+    checkpoint_id: uuid.UUID
+    splitter_options: Optional[Dict[str, Any]] = None
+
```
allensallinger, line 59: Is there a correlation id in this as well of is that only present in the message going from the EventBridge Schedule to the Lambda?

anthonyburdi: Good question - the `correlation_id` is not part of events but passed separately to the backend in the api call (see `_create_scheduled_job_and_set_started`).

Typically the event is created first, then it is sent into the queue for processing - this is when the `correlation_id` is created. Since we are doing the opposite (creating an event from the queue) I chose to pass the `correlation_id` separately to keep the event consistent with the other events.

tests/agent/actions/checkpoints/test_run_checkpoint_action.py
Kilo59: `non-blocking`

Are you using the exact same parameters here and for `test_run_checkpoint_action_raises_on_test_connection_failure`?

If so you put them both in the same [test class](https://docs.pytest.org/en/8.2.x/getting-started.html#group-multiple-tests-in-a-class) and put the parameters at the class level.

anthonyburdi: Oooh good call. Yep - I can clean up either in this PR or a followup.

Shinnnyshinshin (approved):
LGTM! Thanks for the in-person review~

Shinnnyshinshin:
Question: Do we need a new agent release? Will we need to make an additional release this week?

tests/agent/actions/checkpoints/test_run_checkpoint_action.py
anthonyburdi: Not finding a great way to reference as class level variables in the pytest parametrization decorator. I moved them to the module level at least to DRY it up a bit. Open to suggestions for a followup PR if there is a cleaner way.

anthonyburdi:
> Question: Do we need a new agent release? Will we need to make an additional release this week?

Yes we will - I think we can release this during the regularly scheduled release.

PR #301, opened 2024-06-12
Fix Extra dependencies

Kilo59:
The `snowflake` + `postgres` extras were not properly setup for this repo.
```
pip install 'great_expectations_cloud[snowflake]'
```
```
pip install 'great_expectations_cloud[postgres]'
```

The above commands should have installed the dependencies listed in the corresponding poetry extras section but they do not because they must also be set as 'optional' dependencies in our core dependencies section.
https://python-poetry.org/docs/pyproject#extras


poetry.lock
```diff
@@ -4115,10 +4111,10 @@ docs = ["furo", "jaraco.packaging (>=9.3)", "jaraco.tidelift (>=1.4)", "rst.link
 testing = ["big-O", "jaraco.functools", "jaraco.itertools", "more-itertools", "pytest (>=6)", "pytest-checkdocs (>=2.4)", "pytest-cov", "pytest-enabler (>=2.2)", "pytest-ignore-flaky", "pytest-mypy", "pytest-ruff (>=0.2.1)"]
 
 [extras]
-postgres = []
-snowflake = []
```
Kilo59, line 4119: Note, that these were empty before

pyproject.toml
```diff
@@ -31,8 +31,14 @@ orjson = "^3.9.7, !=3.9.10" # TODO: remove inequality once dep resolution issue
 # relying on packaging in agent code so declaring it explicitly here
 packaging = ">=21.3,<25.0"
 tenacity = "^8.2.3"
+# optional dependencies
+snowflake-sqlalchemy = { version = "*", optional = true }
+sqlalchemy = { version = "*", optional = true }
+psycopg2-binary = { version = "*", optional = true }
```
Kilo59, line 37: If we decide we want lower or upper limits we would define these here.

allensallinger (approved):

PR #318, opened 2024-07-02
Remove env var usage from unit tests

anthonyburdi:
We had previously been using environment variables set outside of the unit tests.


tests/integration/conftest.py
```diff
@@ -27,7 +27,10 @@ def token():
 
 
 @pytest.fixture(scope="module")
-def context() -> CloudDataContext:
+def context(monkeypatch, cloud_base_url, org_id, token) -> CloudDataContext:
+    monkeypatch.setenv("GX_CLOUD_ORGANIZATION_ID", org_id)
+    monkeypatch.setenv("GX_CLOUD_ACCESS_TOKEN", token)
+    monkeypatch.setenv("GX_CLOUD_BASE_URL", cloud_base_url)
```
Kilo59, line 33: `GX_CLOUD_ACCESS_TOKEN` and `GX_CLOUD_ORGANIZATION_ID` are still coming from fixtures that are relying on previously set ENV VARs.

Kilo59 (changes requested):
That doesn't actually address the problem of tests relying on prior state.

tests/agent/test_agent.py
```diff
@@ -113,13 +112,13 @@ def local_mercury():
...
-    return os.environ.get("GX_CLOUD_ORGANIZATION_ID")
+def org_id() -> str:
+    return "3ee9791c-4ea6-479d-9b05-98217e70d341"
 
 
 @pytest.fixture
-def token_env_var():
-    return os.environ.get("GX_CLOUD_ACCESS_TOKEN")
+def token() -> str:
+    return "access_token"
```
anthonyburdi, line 121: This is the meat of the change

tests/agent/test_agent.py
```diff
@@ -175,7 +174,7 @@ def test_gx_agent_gets_env_vars_on_init(get_context, gx_agent_config):
     assert agent._config == gx_agent_config
 
 
-def test_gx_agent_invalid_token(monkeypatch):
+def test_gx_agent_invalid_token(monkeypatch, set_required_env_vars: None):
```
anthonyburdi, line 177: Also here, setting other env vars before changing the token to "invalid" on the next line.

tests/agent/test_agent.py
Kilo59: These old fixtures were expressively named, can we make these similar expressive? `dummy_token`, `dummy_org_id`.

Or `random_token`/`random_org_id` (and actually make them random).

anthonyburdi: Good call - I was meaning to keep this PR small but it makes sense to keep all these changes together.

anthonyburdi: Updated

tests/integration/conftest.py
```diff
@@ -16,23 +16,27 @@ def cloud_base_url() -> str:
     return "http://localhost:5000"
 
 
-@pytest.fixture
-def org_id():
-    return os.environ.get("GX_CLOUD_ORGANIZATION_ID")
+@pytest.fixture(scope="module")
+def org_id_env_var() -> str:
```
Kilo59, line 20: Why are these module scoped?

Kilo59 (approved):

tests/integration/conftest.py
anthonyburdi: Good question. Since they are used in the module scoped `context` fixture below. It may be a good idea to convert the `context` fixture (and the fixtures it references) to a function scoped fixture to make sure there are no test interdependencies but I haven't yet had a chance to look at all the tests where it is used to confirm.

PR #323, opened 2024-07-04
Update Pydantic Model - Ignore Extra fields for `EventBase` only

Shinnnyshinshin:
* Followed pattern described [here](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra). 
```python
from pydantic import BaseModel, ConfigDict
class User(BaseModel):
    model_config = ConfigDict(extra='ignore')  

    name: str
``` 

* Model also defaults to `ignore` for extra fields, but there may be value in making this explicit

### DIFF
--- a/great_expectations_cloud/agent/models.py
+++ b/great_expectations_cloud/agent/models.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import uuid
-from typing import Any, Dict, Extra, Literal, Optional, Sequence, Set, Union
+from typing import Any, Dict, Literal, Optional, Sequence, Set, Union
 from uuid import UUID
 
 from great_expectations.compatibility.pydantic import BaseModel, ConfigDict, Field
@@ -12,13 +12,12 @@ from great_expectations_cloud.agent.exceptions import GXCoreError
 
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
-    class Config:
-        # 2024-03-04: ZEL-501 Strictly enforce models for handling outdated APIs
-        extra: str = Extra.forbid
+    model_config = ConfigDict(extra="forbid")
 
 
 class EventBase(AgentBaseModel):
     type: str
+    # ignoring extra fields in Action messages
     model_config = ConfigDict(extra="ignore")
 
 
--- a/great_expectations_cloud/agent/models.py
+++ b/great_expectations_cloud/agent/models.py
@@ -4,7 +4,7 @@ import uuid
 from typing import Any, Dict, Literal, Optional, Sequence, Set, Union
 from uuid import UUID
 
-from great_expectations.compatibility.pydantic import BaseModel, Extra, Field
+from great_expectations.compatibility.pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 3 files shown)


great_expectations_cloud/agent/models.py
```diff
@@ -4,17 +4,15 @@
...
 from typing_extensions import Annotated
 
 from great_expectations_cloud.agent.exceptions import GXCoreError
 
 
 class AgentBaseModel(BaseModel):  # type: ignore[misc] # BaseSettings is has Any type
-    class Config:
-        # 2024-03-04: ZEL-501 Strictly enforce models for handling outdated APIs
-        extra: str = Extra.forbid
+    model_config = ConfigDict(extra="ignore")
```
Kilo59, line 15: This seems dangerous to do this at the level of the `AgentBaseModel`.

Why not do it for the specific models you want this new behavior for?

Or create 2 different `AgentBaseModels`

```
AgentBaseIgnoreExtra
AgentBaseForbitExtra
```

allensallinger: Łukasz had suggested that we do ignore the extra fields here: https://greatexpectationslabs.slack.com/archives/C05R8RE43CL/p1720051367833419?thread_ts=1720046306.191159&cid=C05R8RE43CL

This does make it harder for us to add more fields but it does ensure that we are not accepting in random inputs from messages.

Shinnnyshinshin: Also a helpful thread here: 
https://greatexpectationslabs.slack.com/archives/C04KWBU1XDF/p1718134632349739

Will merge this PR to unblock the work, but happy to talk more next week.

Kilo59 (changes requested):
I really do not think we should be in one go ignoring all fields on all models.
If we want to ignore fields for certain models, lets do that as needed.

great_expectations_cloud/agent/models.py
Kilo59: If we want to ignore fields for certain models please create 2 separate base models and start inheriting from the `AgentBaseIgnoreExtra` where needed but otherwise inherit from `AgentBaseForbidExtra`.

Shinnnyshinshin: I'm against this.  It feel like now is not the right time to put in these classes. Could risk encroaching on over-engineering

Kilo59: Renaming a model is not over-engineering.

Kilo59 (approved):
I still think it would be a good idea to give the AgentBaseModel a more expressive name, and have 2 different versions of it. Especially when working with the runner. But I won't block on that. 
```
AgentBaseIgnoreExtra
AgentBaseForbidExtra
```

Shinnnyshinshin:
> I still think it would be a good idea to give the AgentBaseModel a more expressive name, and have 2 different versions of it. Especially when working with the runner. But I won't block on that.
> 
> ```
> AgentBaseIgnoreExtra
> AgentBaseForbidExtra
> ```

I'd like to revisit this down the line. Happy to take responsibility if it comes back to bite us in a big way 😅

PR #328, opened 2024-07-08
Don't use `gx` pydantic compatibility package. 

Kilo59:
## What

Use pydantic imports directly rather than the `great_expectations.compatiblity.pydantic` package.

Implicit in this change is the decision to stop trying to make `GX-Agent` compatible with every possible environment that `0.18.x` of OSS is. Pydantic v2 is not compatible with some old versions of Airflow for example.

## Why

The GX compatibility layer destroys the typing information of the imported objects, which makes the rest of the code less type safe.
It also abstracts away which version of pydantic is being used, which was a major contributing factor to [the recent incident](https://status.greatexpectations.io/incidents/g509swswhclv)

We can be more opinionated about the environments the agent is compatible with.
Having said that, it seems unlikely that users will want to install/deploy the GX-Agent in their Airflow environment anyways.

## Alernatives

Improve the typing in `great_expectations.compatibility.pydantic` and export types.
- https://github.com/great-expectations/great_expectations/pull/10089


Kilo59:
This change exposes a lot of additional typing issues that need to be addressed or ignored.

great_expectations_cloud/agent/config.py
```diff
@@ -1,19 +1,28 @@
...
+        TODO:once mypy fully support annoting **kwargs with a Unpack[TypedDict], we should do that.
+        https://peps.python.org/pep-0692/
+        """
+        super().__init__(**overrides)
+
 
 def generate_config_validation_error_text(validation_error: ValidationError) -> str:
     missing_variables = ", ".join(
-        [validation_error["loc"][0] for validation_error in validation_error.errors()]
+        [str(validation_error["loc"][0]) for validation_error in validation_error.errors()]
```
Kilo59, line 25: `loc` could also be an `int`.

great_expectations_cloud/agent/actions/agent_action.py
```diff
@@ -3,15 +3,15 @@
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Generic, Sequence, TypeVar
 
-from great_expectations.compatibility.pydantic import BaseModel
+from pydantic.v1 import BaseModel
```
Shinnnyshinshin, line 6: 🚀

pyproject.toml
```diff
@@ -24,7 +24,7 @@ include = [
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
 great-expectations = "^0.18.17"
-pydantic = "<3"
+pydantic = ">=2.8.1,<3"
```
Kilo59, line 27: This now requires pydantic v2.
This soft requirement before is what caused us to use the pydantic compatibility layer which is what led to the recent incident.

tests/integration/actions/test_draft_datasource_config_action.py
```diff
@@ -29,7 +29,7 @@ def test_running_draft_datasource_config_action(
     )
     draft_datasource_config_event = DraftDatasourceConfigEvent(
         type="test_datasource_config",
-        config_id=draft_datasource_id_for_connect_successfully,
+        config_id=UUID(draft_datasource_id_for_connect_successfully),
```
Shinnnyshinshin, line 32: 🚀

Shinnnyshinshin (approved):

PR #339, opened 2024-07-12
Refactor HandleEvent workflow to take in DataContext as parameter

Shinnnyshinshin:
Updating Handle-Event workflow in Agent.
* No behavior is changed, just a helper method is added, and `_handle_event` takes in one additional parameter (`data_context`),  which is then passed to the `EventHandler` rather than `self._data_context` directly. 
* Why is this needed? 
  * GX-Runner inherits from GX-Agent, and requires a DataContext to be retrieved per-event. 
  * This means are able to override the `_get_data_context()` method in the runner repo and keep the event handling behavior consistent between GX-Agent and GX-runner. 
   * [PR in GX-runner that enables it to retrieve a DataContext per event](https://github.com/greatexpectationslabs/gx-runner/pull/58)

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -213,8 +213,11 @@ class GXAgent:
         # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
         self._set_http_session_headers(correlation_id=event_context.correlation_id)
 
-        # send this message to a thread for processing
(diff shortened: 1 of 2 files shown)


anthonyburdi (approved):
LGTM - can you verify locally that you can still run a validation? Also we can verify in dev (since this is a pre-release) as well once this is merged.

Kilo59:
I would suggest any methods that are being overridden or used by the runner be made public.

PR #340, opened 2024-07-12
Add job duration to action result

JennyTee:
* We want to track job duration
* We already have `Starting job` and `Job completed` log events being created in `agent.py`
* These logs can be sent to DataDog
* In this PR we: 1) add org ID to both log event types and 2) add the calculated job duration to the `Job completed` log

Related Slack convo: https://greatexpectationslabs.slack.com/archives/C05R8RE43CL/p1720805454422319

### DIFF
--- a/great_expectations_cloud/agent/actions/agent_action.py
+++ b/great_expectations_cloud/agent/actions/agent_action.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Generic, Optional, Sequence, TypeVar
 
@@ -15,7 +16,7 @@ class ActionResult(BaseModel):
     id: str
     type: str
     created_resources: Sequence[CreatedResource]
-    job_duration_milliseconds: Optional[float] = None  # noqa: UP007  # Python 3.8 doesn't support `X | Y` for type annotations
+    job_duration: Optional[datetime.timedelta] = None  # noqa: UP007  # Python 3.8 doesn't support `X | Y` for type annotation
 
 
 _EventT = TypeVar("_EventT", bound=Event)
--- a/great_expectations_cloud/agent/actions/agent_action.py
+++ b/great_expectations_cloud/agent/actions/agent_action.py
@@ -1,21 +1,12 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Generic, Sequence, TypeVar
+from typing import TYPE_CHECKING, Generic, TypeVar
 
-from pydantic.v1 import BaseModel
-
-from great_expectations_cloud.agent.models import CreatedResource, Event
+from great_expectations_cloud.agent.models import ActionResult, Event
 
 if TYPE_CHECKING:
     from great_expectations.data_context import CloudDataContext
-
-
-class ActionResult(BaseModel):
-    id: str
-    type: str
-    created_resources: Sequence[CreatedResource]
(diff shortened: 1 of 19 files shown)


great_expectations_cloud/agent/event_handler.py
```diff
@@ -72,10 +73,13 @@ def get_event_action(self, event: Event) -> AgentAction[Any]:
         return action_class(context=self._context)
 
     def handle_event(self, event: Event, id: str) -> ActionResult:
```
JennyTee, line 75: I added the duration calculation here inside of the `handle_event` method so that we don't include time the job might spend queued - since I don't think we would want to bill users for that.

allensallinger:
Any log that is emitted from a pod will be scrapped by a DataDog Agent and sent to DataDog's Web UI, except if we filter out the logs like we do with the internal Agents.

At the end of the handle loop we can emit a log. Have the duration field, org_id field and event field on it. With these we will be able to perform a query in DataDog like `service:gx-runner AND event:event_completed` and then do a count of the durations. If we want to see the compute time spent per org_id we can filter on that as well.

great_expectations_cloud/agent/actions/agent_action.py
```diff
@@ -1,22 +1,14 @@
...
 if TYPE_CHECKING:
     from great_expectations.data_context import CloudDataContext
 
 
-class ActionResult(BaseModel):
-    id: str
-    type: str
-    created_resources: Sequence[CreatedResource]
-
-
```
JennyTee, line 19: A lot of the diffs in this file are due to this change--I moved the `ActionResult` model to `models.py`. Reason: I wanted to make the added `job_duration_milliseconds` property optional, but there is a ruff rule that requires us to use `float | None` syntax for this but then python 3.8/9 doesn't support that, so tests were failing.

Moving `ActionResult` to `models.py` means that this issue will be temporarily ignored until we move off of python 3.8/9 and then it'll be easy to address.

The alternative would have been to ignore the ruff issue, but that might have been harder to notice when the time comes to upgrade. The drawback is 1) the large diff here and 2) a larger generic models file.

great_expectations_cloud/agent/actions/agent_action.py
```diff
@@ -15,6 +15,7 @@ class ActionResult(BaseModel):
     id: str
     type: str
     created_resources: Sequence[CreatedResource]
+    job_duration_milliseconds: Optional[float] = None  # noqa: UP007  # Python 3.8 doesn't support `X | Y` for type annotations
```
JennyTee, line 18: This rule is ignored in the models.py file. I could also have moved this type over to that file, but it created a bunch of extra diffs and also makes the models.py file even larger. So I just added this comment instead; we can un-ignore it once we move away from older versions of Python

Kilo59: There's a native `datetime.timedelta` type that better represents this concept, can we use that instead of `float`.
And `pydantic` already knows how to serialize and validate it.

https://docs.python.org/3/library/datetime.html#timedelta-objects

JennyTee: Cool, thanks for the suggestion @Kilo59 . I updated this to be an optional timedelta instead.

For the logger.info mapping, though, I want to convert the timedelta to something easier to report on. Since we expect most jobs to take less than a minute, I went ahead and converted the timedelta to `total_seconds()` in the logs.

Kilo59 (approved):
Looks great!

PR #191, opened 2024-04-08
Add structured JSON agent logs

rreinoldsc:
Structure our logging messages so they can be optionally emitted as JSON

Default:
```
[INFO] great_expectations_cloud.agent.agent: GX Agent version: 0.0.47.dev0
```

JSON:
```
gx-agent --json-log --log-level=info --custom-log-tags='{"environment":"superpowah","service_name":"gx-agent"}'
```
```json
{"event": "GX Agent version: 0.0.47.dev0", "level": "INFO", "logger": "great_expectations_cloud.agent.agent", "service": "gx-agent", "env": "production", "timestamp": ["2024-04-18T11:39:29.901161"]}
```

PR #343, opened 2024-07-15
Add `service_name` to agent env variables and logs

JennyTee:
* Adds service name to the Agent env variables for use in logs
* This is to allow the GXRunner to override the service name to be `gx-runner`

### DIFF
--- a/great_expectations_cloud/agent/config.py
+++ b/great_expectations_cloud/agent/config.py
@@ -3,13 +3,15 @@ from __future__ import annotations
 from great_expectations.data_context.cloud_constants import CLOUD_DEFAULT_BASE_URL
 from pydantic.v1 import AnyUrl, BaseSettings, ValidationError
 
+from great_expectations_cloud.agent.constants import SERVICE_NAME
+
 
 class GxAgentEnvVars(BaseSettings):
     # pydantic will coerce this string to AnyUrl type
     gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
-    gx_cloud_organization_id: str = ""
-    gx_cloud_access_token: str = ""
-    service_name: str = "gx-agent"
+    service_name: str = SERVICE_NAME
+    gx_cloud_organization_id: str
+    gx_cloud_access_token: str
 
     def __init__(self, **overrides: str | AnyUrl) -> None:
         """
--- a/great_expectations_cloud/agent/config.py
+++ b/great_expectations_cloud/agent/config.py
@@ -7,8 +7,8 @@ from pydantic.v1 import AnyUrl, BaseSettings, ValidationError
 class GxAgentEnvVars(BaseSettings):
     # pydantic will coerce this string to AnyUrl type
     gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
-    gx_cloud_organization_id: str
-    gx_cloud_access_token: str
+    gx_cloud_organization_id: str = ""
+    gx_cloud_access_token: str = ""
     service_name: str = "gx-agent"
 
     def __init__(self, **overrides: str | AnyUrl) -> None:
--- a/great_expectations_cloud/agent/config.py
+++ b/great_expectations_cloud/agent/config.py
@@ -9,6 +9,7 @@ class GxAgentEnvVars(BaseSettings):
     gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
     gx_cloud_organization_id: str
     gx_cloud_access_token: str
+    service_name: str = "gx-agent"
 
     def __init__(self, **overrides: str | AnyUrl) -> None:
         """
--- a/great_expectations_cloud/agent/constants.py
+++ b/great_expectations_cloud/agent/constants.py
@@ -9,6 +9,7 @@ class HeaderName(str, enum.Enum):
     AGENT_JOB_ID = "Agent-Job-Id"
 
 
-USER_AGENT_HEADER: Final = "gx-agent"
+USER_AGENT_HEADER: Final[str] = "gx-agent"
+SERVICE_NAME: Final[str] = "gx-agent"
 
 __all__ = ["USER_AGENT_HEADER"]
--- a/great_expectations_cloud/logging/logging_cfg.py
+++ b/great_expectations_cloud/logging/logging_cfg.py
@@ -12,13 +12,13 @@ from typing import Any, ClassVar, Final, Literal
 
 from typing_extensions import override
 
-from great_expectations_cloud.agent import GXAgent
+from great_expectations_cloud.agent.config import GxAgentEnvVars
 
 LOGGER = logging.getLogger(__name__)
-
+env_vars = GxAgentEnvVars()
 DEFAULT_LOG_FILE: Final[str] = "logfile"
 DEFAULT_LOG_DIR = "logs"
-SERVICE_NAME: Final[str] = GXAgent.get_config().service_name
+SERVICE_NAME: Final[str] = env_vars.service_name
 DEFAULT_FILE_LOGGING_LEVEL: Final[int] = logging.DEBUG
 
 # Consider moving to file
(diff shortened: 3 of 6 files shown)


great_expectations_cloud/agent/config.py
```diff
@@ -7,8 +7,9 @@
 class GxAgentEnvVars(BaseSettings):
     # pydantic will coerce this string to AnyUrl type
     gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL  # type: ignore[assignment]
-    gx_cloud_organization_id: str
-    gx_cloud_access_token: str
+    gx_cloud_organization_id: str = ""
+    gx_cloud_access_token: str = ""
```
JennyTee, line 11: I could use a second set of eyes on this. Without adding the default values, `test_logging_config` was failing with this error:

 ```
______________________________________________ ERROR collecting tests/test_logging_cfg.py _______________________________________________
tests/test_logging_cfg.py:13: in <module>
    from great_expectations_cloud.logging.logging_cfg import (
great_expectations_cloud/logging/logging_cfg.py:21: in <module>
    SERVICE_NAME: Final[str] = GxAgentEnvVars().service_name
great_expectations_cloud/agent/config.py:21: in __init__
    super().__init__(**overrides)
.venv/lib/python3.10/site-packages/pydantic/v1/env_settings.py:40: in __init__
    super().__init__(
.venv/lib/python3.10/site-packages/pydantic/v1/main.py:341: in __init__
    raise validation_error
E   pydantic.v1.error_wrappers.ValidationError: 2 validation errors for GxAgentEnvVars
E   gx_cloud_organization_id
E     field required (type=value_error.missing)
E   gx_cloud_access_token
E     field required (type=value_error.missing)
```

I wasn't sure how else to fix this

Comment (Shinnnyshinshin) [reply great_expectations_cloud/agent/config.py:11]: yea we see this in the Runner code as well. There we use `monkeypatch` to set values to the env vars.  @anthonyburdi  I know you worked on ENV vars and tests recently.. is this a viable approach here too?

https://github.com/greatexpectationslabs/gx-runner/blob/dd2334ef3a5f1bc885c9827a75e867e53adfd434/tests/test_runner.py#L55

Comment (anthonyburdi) [reply great_expectations_cloud/agent/config.py:11]: Good call @Shinnnyshinshin - yes we used `monkeypatch` to set these env variables. We don't want to rely on the state of the environment for unit tests. Take a look at this PR for some examples: https://github.com/great-expectations/cloud/pull/318/files. You may want to create a fixture like the `set_required_env_vars` fixture to use in all the tests in `tests/test_logging_cfg.py`

File: great_expectations_cloud/logging/logging_cfg.py [hunk 2]
```diff
@@ -12,11 +12,13 @@
...
 from typing_extensions import override
 
+from great_expectations_cloud.agent.config import GxAgentEnvVars
+
 LOGGER = logging.getLogger(__name__)
 
 DEFAULT_LOG_FILE: Final[str] = "logfile"
 DEFAULT_LOG_DIR = "logs"
-SERVICE_NAME: Final[str] = "gx-agent"
+SERVICE_NAME: Final[str] = GxAgentEnvVars().service_name
```
Kilo59, line 21: I would flip this dependency if possible.
Make the default `GxAgentEnvVars.service_name` come from a constant, rather than this constant coming from `GxAgentEnvVars().service_name`.

great_expectations_cloud/agent/config.py
Kilo59: Yes I would use `monkeypatch.setenv` for this.
And ideally create a fixture that sets all the needed env vars.

https://docs.pytest.org/en/4.6.x/monkeypatch.html#monkeypatching-environment-variables

great_expectations_cloud/logging/logging_cfg.py
JennyTee: Thanks - I added the default value to the existing `great_expectations_cloud/agent/constants.py ` file.

Because I need the runner to be able to override this value, I still need to keep it as part of `GxAgentEnvVars` and then use the env vars to get the SERVICE_NAME in this file. I don't know exactly how it works, but the SERVICE_NAME is added to the logger log event. Here's an example log pulled from ArgoCD:
```
{
  "args": [],
  "levelname": "INFO",
  "filename": "agent.py",
  "module": "agent",
  "exc_info": null,
  "stack_info": null,
  "funcName": "_handle_event_as_thread_exit",
  "created": 1720803645.03965,
  "relativeCreated": 154868265.17105103,
  "event_type": "run_scheduled_checkpoint.received",
  "correlation_id": "a01d7125-a5b7-4791-8e76-198a01958551",
  "event": "Completed job",
  "level": "INFO",
  "logger": "great_expectations_cloud.agent.agent",
  "timestamp": "2024-07-12T17:00:45.039650+00:00",
  "environment": "production",
  "service_name": "gx-agent"
}
```

great_expectations_cloud/agent/config.py
JennyTee: Thanks, all. I have been trying different variations of `monkeypatch` but am having issues b/c the tests are erroring on the `logging_cfg` import, and I can't figure out how to mock this sort of import.

great_expectations_cloud/logging/logging_cfg.py
```diff
@@ -12,11 +12,15 @@
...
 
+from great_expectations_cloud.agent.config import GxAgentEnvVars
+
 LOGGER = logging.getLogger(__name__)
 
 DEFAULT_LOG_FILE: Final[str] = "logfile"
 DEFAULT_LOG_DIR = "logs"
-SERVICE_NAME: Final[str] = "gx-agent"
+env_vars = GxAgentEnvVars()
+SERVICE_NAME = env_vars.service_name
```
Kilo59, line 22: Where does this `logging_cfg.SERVICE_NAME` variable actually get used?

I would not define a module level variable here at all if possible and instead pull the `GxAgentEnvVars().service_name` when its actually needed.

great_expectations_cloud/agent/constants.py
```diff
@@ -9,6 +9,7 @@ class HeaderName(str, enum.Enum):
     AGENT_JOB_ID = "Agent-Job-Id"
 
 
-USER_AGENT_HEADER: Final = "gx-agent"
+USER_AGENT_HEADER: Final[str] = "gx-agent"
+SERVICE_NAME: Final[str] = "gx-agent"
 
 __all__ = ["USER_AGENT_HEADER"]
```
Kilo59, line 15:
```suggestion
__all__ = ["USER_AGENT_HEADER", "SERVICE_NAME"]
```

great_expectations_cloud/logging/logging_cfg.py
Kilo59: This sounds like a symptom of using a module level variable.

Where does this `logging_cfg.SERVICE_NAME` variable actually get used?

I would not define a module level variable here at all if possible and instead pull the `GxAgentEnvVars().service_name` when its actually needed.

great_expectations_cloud/agent/config.py
JennyTee: Thanks!

> Where does this logging_cfg.SERVICE_NAME variable actually get used?

I was super confused about this; thought it was some magic variable name or something that was causing the service name to show in the logs.

Looking at this more closely, though, it looks like we *not* using this variable at all. Instead, we are passing in a custom log tag for `service_name` in the Kubernetes manifest. 

So TL;DR, it looks ike none of these changes are actually needed. When we run the GX Runner agent, we can just update the manifest to use `gx-runner` as the `service_name` and then the logs pushed to Datadog will include the appropriate service name.

I'll close out this PR. Thanks again for the feedback, much appreciated.
cc @allensallinger

Kilo59: FWIW, I do think it would be valuable for a user to be able to control the service name (via env var or command line argument).

An organization likely may need to deploy multiple agents to connect to all of its different data sources (or would want to for security reasons).
In that case, being able to customize the `service-name` can be helpful to distinguish them.

JennyTee: The agent cli accepts an argument for `--custom-log-tags` that @rreinoldsc added in #191 , so we should be good there!


https://github.com/great-expectations/cloud/blob/f5ffaab324ca770c8be2303843b7baecb2e5b950/great_expectations_cloud/agent/cli.py#L56

PR #346, opened 2024-07-16
Refactor `set_http_session_headers` to be compatible with GX-runner

Shinnnyshinshin:


great_expectations_cloud/agent/agent.py
```diff
@@ -118,17 +118,17 @@ def __init__(self: Self):
         print(f"GX Agent version: {agent_version}")
         print(f"Great Expectations version: {great_expectations_version}")
         print("Initializing the GX Agent.")
-        self._set_http_session_headers()
```
Shinnnyshinshin, line 121: refactored to take in `data_context` as parameter and moved it past the `self._context` initialization. behavior should be the same, but allows GX-runner to pass in a DataContext-per job. 

https://github.com/greatexpectationslabs/gx-runner/pull/74

great_expectations_cloud/agent/agent.py
```diff
@@ -210,13 +210,16 @@ def _handle_event_as_thread_enter(self, event_context: EventContext) -> None:
             self._redeliver_msg_task = loop.create_task(event_context.redeliver_message())
             return
 
+        data_context = self.get_data_context(event_context=event_context)
         # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
-        self._set_http_session_headers(correlation_id=event_context.correlation_id)
+        self._set_http_session_headers(
```
Shinnnyshinshin, line 215: This is the key change : when the Agent enters a job thread, it runs `get_data_context()` to return either `self._data_context` in the agent, or the job-DataContext in the GX-runner.  

The  returned DataContext is then used to `_set_http_session_headers()` for the current job.

anthonyburdi (approved):
LGTM

great_expectations_cloud/agent/agent.py
Kilo59: You'll also want to be able to change the `HeaderName.USER_AGENT` value on line 472/402 so that we can distinguish between Agent calls and runner class (and the versions used).

anthonyburdi (changes requested):
Just need to address @Kilo59's comment and also change the release version after the agent release.

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: That's a great point @Kilo59... with the dependency on the values in `constants.py` and since we are generating a new log message and header, it might be worth considering just overriding this method in the `gx-runner`. Can we discuss @anthonyburdi ?

anthonyburdi (approved):

great_expectations_cloud/agent/agent.py
anthonyburdi: @Kilo59 We ended up using overridable methods to get `HeaderName` and `USER_AGENT_HEADER` so that in the runner we can pull from a new constants.py file, see: https://github.com/greatexpectationslabs/gx-runner/pull/74
We left the header name `Agent-Job-Id` to not mess with downstream processing (mercury). If that is not a concern, happy to revisit.

PR #348, opened 2024-07-17
sqlalchemy 2 - stricter dependency constraints 

Kilo59:
Require SQLAlchemy 2 and stop using great_expectations compatibility layer.


Kilo59:
- done as part of #368

PR #349, opened 2024-07-17
Use overridable methods for org ID and auth token

anthonyburdi:
The runner needs to retrieve these values differently. Adding two small methods we can override in the runner to keep the code consistent. We can diverge later if needed.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -10,6 +10,7 @@ from concurrent.futures.thread import ThreadPoolExecutor
 from functools import partial
 from importlib.metadata import version as metadata_version
 from typing import TYPE_CHECKING, Any, Callable, Dict, Final
+from uuid import UUID
 
 import orjson
 from great_expectations import get_context  # type: ignore[attr-defined] # TODO: fix this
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -400,10 +416,9 @@ def _update_status(self, job_id: str, status: JobStatus) -> None:
         """
         LOGGER.info("Updating status", extra={"job_id": job_id, "status": str(status)})
         agent_sessions_url = (
-            f"{self._config.gx_cloud_base_url}/organizations/{self._config.gx_cloud_organization_id}"
-            + f"/agent-jobs/{job_id}"
+            f"{self._config.gx_cloud_base_url}/organizations/{org_id}" + f"/agent-jobs/{job_id}"
```
Shinnnyshinshin, line 419: do we need to convert to `str()` here?

great_expectations_cloud/agent/agent.py
```diff
@@ -425,10 +440,10 @@ def _create_scheduled_job_and_set_started(self, event_context: EventContext) ->
         LOGGER.info("Creating scheduled job and setting started", extra=data)
 
         agent_sessions_url = (
-            f"{self._config.gx_cloud_base_url}/organizations/{self._config.gx_cloud_organization_id}"
+            f"{self._config.gx_cloud_base_url}/organizations/{self.get_organization_id()}"
```
Shinnnyshinshin, line 443: also check if this needs to be `str()` ? Not 100% sure

Shinnnyshinshin: confirmed this is ok.

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: nope. confirmed this is ok

Shinnnyshinshin (approved):

PR #350, opened 2024-07-17
Pass event context

anthonyburdi:
Followup to https://github.com/great-expectations/cloud/pull/349 to pass the event context for retrieval of the organization ID in the runner.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -231,10 +231,10 @@ class GXAgent:
             self._current_task.add_done_callback(on_exit_callback)
 
     def get_data_context(self, event_context: EventContext) -> CloudDataContext:
-        """Helper method to get a DataContext Agent. Overriden in GX-Runner."""
+        """Helper method to get a DataContext Agent. Overridden in GX-Runner."""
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -440,7 +440,7 @@ def _create_scheduled_job_and_set_started(self, event_context: EventContext) ->
         LOGGER.info("Creating scheduled job and setting started", extra=data)
 
         agent_sessions_url = (
-            f"{self._config.gx_cloud_base_url}/organizations/{self.get_organization_id()}"
+            f"{self._config.gx_cloud_base_url}/organizations/{self.get_organization_id(event_context)}"
```
allensallinger, line 443: Does the UUID get converted to str automatically here?

allensallinger (approved):

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: @allensallinger good question. we were able to confirm that yes it does

Shinnnyshinshin (approved):

PR #351, opened 2024-07-17
Add processed_by field to JobCompleted event

JennyTee:
Adds and populates `processed_by` field to JobCompleted event so we can report on jobs completed by the agent vs. the runner.

⚠️  https://github.com/greatexpectationslabs/mercury/pull/3575 should be merged before this change.


great_expectations_cloud/agent/agent.py
```diff
@@ -299,6 +299,7 @@ def _handle_event_as_thread_exit(
                     success=False,
                     created_resources=[],
                     error_stack_trace="The version of the GX Agent you are using does not support this functionality. Please upgrade to the most recent image tagged with `stable`.",
+                    processed_by="runner" if self._config.queue == "gx-runner" else "agent",
```
JennyTee, line 302: This seemed like the easiest way to accomplish adding `processed_by` here. I'm open to other ideas, though, especially if we don't want the agent to know about the runner queue.

great_expectations_cloud/agent/agent.py
```diff
@@ -343,6 +345,10 @@ def _handle_event_as_thread_exit(
         event_context.processed_successfully()
         self._current_task = None
 
+    def _get_processed_by(self) -> str:
+        """Return the name of the service that processed the event."""
+        return "runner" if self._config.queue == "gx-runner" else "agent"
+
```
JennyTee, line 351: This seemed like the easiest way to accomplish adding processed_by here. I'm open to other ideas, though, especially if we don't want the agent to know about the runner queue.

shiplet:
Only question is whether we can default this repo's event to `processed_by=agent` because we have a gx-runner repo where we can default to `processed_by=runner`? Or do we need to have the ternary split here until the runner repo is fully launched?

anthonyburdi (approved):

JennyTee:
> Only question is whether we can default this repo's event to `processed_by=agent` because we have a gx-runner repo where we can default to `processed_by=runner`? Or do we need to have the ternary split here until the runner repo is fully launched?

So from my understanding b/c the `GXRunner` class extends `GXAgent` (see line [here](https://github.com/greatexpectationslabs/gx-runner/blob/289a30b89cb18626c3d81705ffdb0fa4bc61743d/gx_runner/runner.py#L45)), it's the Agent code that is getting executed when the `JobCompleted` event is built out. 

In my mind the options to populate the `processed_by` field are either:
1. Make the function that builds the event overridable so that the GXRunner can specify another implementation (imo this is not the best b/c the only difference is the `processed_by` field value)
2. Have the GXRunner pass a runner-specific value to the Agent (this is what it's currently doing via the `GXAgentConfig` object).

I think we could theoretically add a `processed_by_name` or `service_name` to the `GXAgentConfig` object and use that value as `processed_by`. But since we already have the `queue` name, I just used that b/c it was easier. But perhaps it is a better long-term solution to not make the Agent aware of the Runner.

PR #352, opened 2024-07-17
Pass in base_url, org ID and auth key to each action

anthonyburdi:
Pass in information to each action that is needed to make calls back to cloud. Formerly these were retrieved from env vars but these need to be retrieved differently in the runner vs the agent so are abstracted out into the previously added functions `get_organization_id`, `get_data_context` and base URL is still retrieved from env var.

### DIFF
--- a/great_expectations_cloud/agent/actions/run_metric_list_action.py
+++ b/great_expectations_cloud/agent/actions/run_metric_list_action.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
+from uuid import UUID
 
 from great_expectations.experimental.metric_repository.batch_inspector import (
     BatchInspector,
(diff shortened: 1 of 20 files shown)


tests/agent/actions/test_draft_datasource_config.py
```diff
@@ -121,7 +127,13 @@ def test_test_draft_datasource_config_success_sql_ds(
     mock_inspector.get_table_names.return_value = table_names
 
     env_vars = GxAgentEnvVars()
-    action = DraftDatasourceConfigAction(context=mock_context)
+    org_id = UUID("81f4e105-e37d-4168-85a0-2526943f9956")
+    action = DraftDatasourceConfigAction(
+        context=mock_context,
+        base_url=" https://api.greatexpectations.io/",
```
Shinnnyshinshin, line 133: is the space intentional?

great_expectations_cloud/agent/actions/run_metric_list_action.py
```diff
@@ -32,13 +33,18 @@ class MetricListAction(AgentAction[RunMetricsListEvent]):
     # TODO: New actions need to be created that are compatible with GX v1 and registered for v1.
     #  This action is registered for v0, see register_event_action()
 
-    def __init__(
+    def __init__(  # noqa: PLR0913  # Refactor opportunity
```
Shinnnyshinshin, line 36: :)

Shinnnyshinshin (approved):
Let's gooo

PR #353, opened 2024-07-18
Ruff `0.5.3`

Kilo59:
Update to `ruff` `0.5.3` and enable Github Actions annotations on linter errors.

https://docs.astral.sh/ruff/integrations/#github-actions

image


tasks.py
```diff
@@ -70,6 +71,8 @@ def lint(ctx: Context, check: bool = False, unsafe_fixes: bool = False) -> None:
     cmds = ["ruff", "check", "."]
     if not check:
         cmds.append("--fix")
+    elif os.getenv("GITHUB_ACTIONS"):
+        cmds.append("--output-format=github")
```
Kilo59, line 75: If this env var is present we are running inside the Github Actions runner and should use the GithubAction output format.

anthonyburdi (approved):
🚀

PR #356, opened 2024-07-19
WIP Make Agent Multi Threaded

anthonyburdi:
Some notes for making the Agent (and thus Runner) multi-threaded.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -125,6 +125,7 @@ class GXAgent:
         with warnings.catch_warnings():
             # suppress warnings about GX version
             warnings.filterwarnings("ignore", message="You are using great_expectations version")
+            # TODO: Instantiate the context in the sub thread (_handle_event()) instead of here.
             self._context: CloudDataContext = get_context(cloud_mode=True)
         print("DataContext is ready.")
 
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -134,8 +135,15 @@ def __init__(self: Self):
         # GX processes and maintain our connection to the broker. Note that
         # the CloudDataContext cached here is used by the worker, so
         # it isn't safe to increase the number of workers running GX jobs.
+        # TODO: Change to use ThreadPoolExecutor max_workers set via environment variable
+        #   with a reasonable default (maybe 1 for the agent at first)
```
Kilo59, line 139: What are we trying to achieve by making the agent multi-threaded?

If we are going for concurrency + parallelism, the more natural path would be to use multi-processing. Using multi-processing means we don't have to worry about a whole class of problems.
1. Thread-unsafe operations no longer become an issue
2. Leaking of state in a multi-tenant environment (while not impossible, becomes much harder to do unintentionally). 


However, I would also strongly suggest we take a step back from introducing this additional complexity and consider adopting a dedicated framework which can make it much easier to achieve our aims while avoiding a lot of the common pitfalls. 

https://faststream.airt.ai/latest/faststream/

wookasz: @Kilo59 how much of an effort do you think moving to faststream would be?

Kilo59: Currently I think very low due to the low complexity of our application.
A few days to 2 weeks.

Admittedly, I've never actually used it [outside of this POC](https://github.com/greatexpectationslabs/REST-V2/pull/7/files#diff-e0eb8b541fb2181fa04bc70ed2f88569b51a5f5f83d570c3def6106e11d24262).
But the concepts are very familiar to me.

If this is something we want to pursue I would suggest we first do a (1 day timeboxed) spike specifically on 2 things.
1. Using Faststream where the producer code lives in a separate repo.
2. Using RabbitMQ with FastStream, specifically with how our Agent + Runner exchanges are set up.

I don't anticipate either of these will be a problem, but I wouldn't want to dive into a project like this without verifying this 2 up front either.

anthonyburdi:
Currently planning to use faststream instead.

PR #359, opened 2024-07-23
Get the context and set http headers in the inner event loop.

anthonyburdi:
Move `get_context()` and `_set_http_session_headers()` into the inner event loop. This is to help make sure action execution is isolated.


great_expectations_cloud/agent/agent.py
```diff
@@ -254,6 +245,13 @@ def _handle_event(
             event_context: event with related properties and actions.
         """
         # warning:  this method will not be executed in the main thread
+
+        data_context = self.get_data_context(event_context=event_context)
+        # ensure that great_expectations.http requests to GX Cloud include the job_id/correlation_id
```
Shinnnyshinshin, line 250: 👍🏼

great_expectations_cloud/agent/agent.py
```diff
@@ -499,7 +500,7 @@ def _set_http_session_headers(
             )
             return
 
-        agent_version = self.get_current_gx_agent_version()
+        agent_version = self._get_version()
```
Shinnnyshinshin, line 503: 🚀

Shinnnyshinshin (approved):

great_expectations_cloud/agent/agent.py
```diff
@@ -468,6 +466,9 @@ def get_header_name(self) -> type[HeaderName]:
     def get_user_agent_header(self) -> str:
         return USER_AGENT_HEADER
 
+    def _get_version(self) -> str:
+        return self.get_current_gx_agent_version()
```
Kilo59, line 470:
```suggestion
    def get_version(self) -> str:
        return self.get_current_gx_agent_version()
```

Kilo59 (approved):

PR #361, opened 2024-07-24
allow disabling progress bars when setting up the agent

TrangPham:
Allows the progress bars to be disabled when GXAgent is initialized

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "20240723.0"
+version = "20240724.0.dev0"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 repository = "https://github.com/great-expectations/cloud"


great_expectations_cloud/agent/agent.py
```diff
@@ -113,7 +114,7 @@ class GXAgent:
     _PYPI_GX_AGENT_PACKAGE_NAME = "great_expectations_cloud"
     _PYPI_GREAT_EXPECTATIONS_PACKAGE_NAME = "great_expectations"
 
-    def __init__(self: Self):
+    def __init__(self: Self, enable_progress_bars: bool = True) -> None:
```
Kilo59, line 117: What's the mechanism to actually disable this?
A CLI argument?
Or is the plan for the GX-Runner to disable this by default?

Kilo59 (approved):

great_expectations_cloud/agent/agent.py
```diff
@@ -126,6 +127,15 @@ def __init__(self: Self):
             # suppress warnings about GX version
             warnings.filterwarnings("ignore", message="You are using great_expectations version")
             self._context: CloudDataContext = get_context(cloud_mode=True)
+            if not enable_progress_bars:
```
anthonyburdi, line 130: This LGTM - just a heads up that this needs to be done in get_data_context() in the runner (since the context is created for each run).

great_expectations_cloud/agent/agent.py
anthonyburdi: The runner overrides `__init__()` so would need to be handled differently there.

anthonyburdi (approved):
LGTM - just a couple of comments about the runner.

PR #366, opened 2024-07-29
use with constructor for sessions

shiplet:
Noticed that each session allocated additional memory without also deallocating it.

This leverages Python's `with` constructor to ensure the Requests library is able to correctly manage memory for the sessions it creates.


Shinnnyshinshin:
@shiplet I had a chance to play around with the tests in `test_agent.py`, and you're absolutely right about the `create_session` preventing it from being mocked properly.

I was able to get it working with a `requests_post`, which does the same mock but in the `post` method that is called by the session.  It allows us to more-directly patch the method that is called. 

Could you give it a try? happy to pair on it tomorrow as well.

```python

@pytest.fixture(autouse=True)
def requests_post(mocker, queue, connection_string):
    """Patch for requests.Session.post"""
    requests_post = mocker.patch("requests.Session.post")
    requests_post().json.return_value = {
        "queue": queue,
        "connection_string": connection_string,
    }
    requests_post().ok = True
    return requests_post

def test_gx_agent_gets_env_vars_on_init(get_context, gx_agent_config, requests_post):
    agent = GXAgent()
    assert agent._config == gx_agent_config

```

Screenshot 2024-07-29 at 7 43 15 PM

shiplet:
@Shinnnyshinshin Great suggestion 🙇 . That seems to have resolved all failures except for `test_gx_agent_updates_cloud_on_job_status` and `test_gx_agent_sends_request_to_create_scheduled_job`. I'll check those out tomorrow AM.

PR #370, opened 2024-07-31
Remove all but standardized version of python in build

wookasz:
Remove usage of python 3.8 in CI steps and all 3.8, 3.9, and 3.11 testing

## Why
We're only using 3.10.14 in our `docker` image and we're hitting up against GitHub rate limits

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.10.14"]
     runs-on: ubuntu-latest
 
     steps:
@@ -83,7 +83,7 @@ jobs:
   integration-tests:
     strategy:
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.10.14"]
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.10.14"]
```
Kilo59, line 45: This has come up before and I think it was decided to keep testing all versions.
However python 3.8 is going to EOL in 3 months so I think it's a relatively easy call to drop support for that.
https://endoflife.date/python

Even if product said we can drop support for everything but `3.10` I would still suggest we keep testing (at least unit-tests) on 3.11 and higher to keep our upgrade path clear.

Having said all that.
It would also be trivial to deal with the github API rate limit by making that test only run on one of python versions.

cc @Shinnnyshinshin @annabarr 
```suggestion
        python-version: ["3.9", "3.10.14", "3.11"]
```

wookasz: @Kilo59 can you point me to where this was decided?

wookasz: I don't understand why we would add unnecessary complexity for ourselves if we don't support running this outside of docker

Shinnnyshinshin: @wookasz is it possible to know the version that people are running when downloading the `agent` from pypi?

Kilo59: @Shinnnyshinshin @wookasz 
https://pypistats.org/packages/great-expectations-cloud

wookasz:
closing stale PR

PR #375, opened 2024-08-05
add org_id to log payload & cleanup

shiplet:
# What
This standardizes a couple things around `organization_id`. Chiefly:
- It's now passed as part of the `extra` payload for each call to `LOGGER`, rather than just the errors.
- It's no longer passed as an internal function argument, instead defaulting to pulling the value from the class properties.
- It shortens the variable names from `organization_id` to `org_id` because it's more concise.

It also removes unused `event_context` parameters from the function signatures of `get_data_context` and `get_organization_id`.

# Why
As we're building out additional features around connecting to and automating connections to data sources, we'll need to increase and standardize visibility into the intended actions and their originating organizations.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -464,7 +464,7 @@ class GXAgent:
             payload = Payload(data=data)
             session.post(agent_sessions_url, data=payload.json())
             LOGGER.info(
-                "Created scheduled job and set started", extra={**data, "organization_id": org_id}
+                "Created scheduled job and set started", extra={**data, "organization_id": str(org_id)}
             )
 
     def get_header_name(self) -> type[HeaderName]:
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -454,7 +454,8 @@ class GXAgent:
             "event": event_context.event.dict(),
         }
         LOGGER.info(
-            "Creating scheduled job and setting started", extra={**data, "organization_id": org_id}
+            "Creating scheduled job and setting started",
+            extra={**data, "organization_id": str(org_id)},
         )
 
         agent_sessions_url = (
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -464,7 +464,8 @@ class GXAgent:
             payload = Payload(data=data)
             session.post(agent_sessions_url, data=payload.json())
             LOGGER.info(
-                "Created scheduled job and set started", extra={**data, "organization_id": str(org_id)}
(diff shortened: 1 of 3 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -246,29 +246,27 @@ def _handle_event(self, event_context: EventContext) -> ActionResult:
         """
         # warning:  this method will not be executed in the main thread
 
-        data_context = self.get_data_context(event_context=event_context)
```
Shinnnyshinshin, line 249: Trying to reconcile this change with needing to have a DataContext-per-Event in the Runner. How does `get_data_context()` handle it?

rreinoldsc (approved):

great_expectations_cloud/agent/agent.py
```diff
@@ -444,20 +448,24 @@ def _create_scheduled_job_and_set_started(self, event_context: EventContext) ->
...
-            f"{self._config.gx_cloud_base_url}/organizations/{self.get_organization_id(event_context)}"
-            + "/agent-jobs"
+            f"{self._config.gx_cloud_base_url}/organizations/{org_id}" + "/agent-jobs"
         )
         with create_session(access_token=self.get_auth_key()) as session:
             payload = Payload(data=data)
             session.post(agent_sessions_url, data=payload.json())
-            LOGGER.info("Created scheduled job and set started", extra=data)
+            LOGGER.info(
+                "Created scheduled job and set started", extra={**data, "organization_id": org_id}
```
rreinoldsc, line 467:
```suggestion
                "Created scheduled job and set started", extra={**data, "organization_id": str(org_id)}
```

rreinoldsc: This `**data` might actually barf in `logging` if values arent serializable

rreinoldsc: Also a bit risky to log without knowing sensitivity of data

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: just read the description more closely :) makes sense

great_expectations_cloud/agent/agent.py
shiplet: Thanks!

great_expectations_cloud/agent/agent.py
shiplet: Awesome! Yeah from the looks of it, the Runner is indeed using the event_context on each call to `get_data_context`, but the Agent isn't. Python also doesn't really care about overridden class methods having matching function signatures, so it _should_ be safe to delete. I'll run some tests to make sure though.

PR #381, opened 2024-08-05
Check matching Org ID between Action and Handler

Shinnnyshinshin:
Prevent the running of any `Events` where the organization ID is different than the one in the Context or handler. 

*  This prevents processing events that are not intended for the current organization, and potentially leaking sensitive information across organizations.


great_expectations_cloud/agent/agent.py
```diff
@@ -550,9 +551,3 @@ def _update_headers_agent_patch(
         # TODO: this is relying on a private implementation detail
         # use a public API once it is available
         http._update_headers = _update_headers_agent_patch
-
-
-class GXAgentError(Exception): ...
-
-
-class GXAgentConfigError(GXAgentError): ...
```
Shinnnyshinshin, line 558: Moving to `exceptions` class to prevent circular-import when using error in `EventHandler`.

shiplet (approved):

great_expectations_cloud/agent/event_handler.py
```diff
@@ -67,6 +68,11 @@ def get_event_action(
         self, event: Event, base_url: str, auth_key: str, organization_id: UUID
     ) -> AgentAction[Any]:
         """Get the action that should be run for the given event."""
+
+        if not self._check_event_organization_id(event, organization_id):
+            # Making message more generic
```
Shinnnyshinshin, line 73: take this out before merging

PR #384, opened 2024-08-07
Add `snowflake-connector-python` to optional dependency

Shinnnyshinshin:


pyproject.toml
```diff
@@ -34,6 +34,7 @@ tenacity = ">=8.2.3,<10.0.0"
 sqlalchemy = { version = ">=2.0", optional = false }
 # optional dependencies
 snowflake-sqlalchemy = { version = ">=1.6", optional = true }
+snowflake-connector-python = { version = ">=3.3.1", optional = true }
```
Shinnnyshinshin, line 37: It was not added to the optional dependencies section, so the versioning wasn't being enforced.

shiplet (approved):

PR #393, opened 2024-08-14
Prevent `pypi` rate-limiting in tests.

Kilo59:
Prevent getting rate limited by the pypi API.

Running this test involves hitting the `pypi` API for version information. When we run it for every version of Python in parallel, we can occasionally hit the rate limit and fail the CI.

This change skips this test for every minor version of python except the one declared in the `.python-version` file.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,7 +70,7 @@ jobs:
         run: poetry install --sync
 
       - name: Run tests
-        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml
+        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml -rEfxXs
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         run: poetry install --sync
 
       - name: Run tests
-        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml
+        run: poetry run pytest -m "not agentjobs and not integration" --cov-report xml -ra
```
Kilo59, line 73: This adds a summary report that shows what tests (if any) had a status other than `pass`.
https://docs.pytest.org/en/latest/how-to/output.html#producing-a-detailed-summary-report

shiplet (approved):
Makes sense to me!

PR #397, opened 2024-08-15
Close connections in datasources

anthonyburdi:


great_expectations_cloud/agent/actions/draft_datasource_config_action.py
```diff
@@ -63,6 +64,11 @@ def check_draft_datasource_config(
             table_names = self._get_table_names(datasource=datasource)
             self._update_table_names_list(config_id=event.config_id, table_names=table_names)
 
+        # Close connections that were opened to avoid running out of connections.
+        execution_engine = datasource.get_execution_engine()
+        if isinstance(execution_engine, SqlAlchemyExecutionEngine):
+            execution_engine.close()
```
Kilo59, line 70: This won't be enough to close all the sqlalchemy engine connections.
Some of our datasource use 2 different sqlalchemy engine instances.

Lets chat about this soon, but I think we need to add a `datasource.close_connection()`/`datasource.shutdown_connection()` method to the `Datasource` interface and let each datasource implement the respective close connection operation.

anthonyburdi: Oh interesting. By two different engine instances do you mean the `backup_engine` in the execution engine?

Kilo59: No.
See the description of this abandoned PR.
But TLDR the SQLALchemy `engine` used in `datasource.test_connection()` is not the same engine used by the `SQLAlchemyExecutionEngine`. 

As described 👇 this has been fixed for SnowflakeDatasource, but it's still a problem for `Postgres`, `DatabricksSQL` and every other SQL datasource.

- https://github.com/great-expectations/great_expectations/pull/9051

anthonyburdi: Oh interesting. I did not know about the duplicate engine issue.
I also am realizing now that I missed that `execution_engine.close()` does not close the actual connection if the `_engine_backup` doesn't exist. I'm going to modify this PR to also explicitly call close on the connection just to test if it helps, but it sounds like we have more work to do overall.

tyler-hoffman:
4 months old and merge conflicts - closing

PR #407, opened 2024-08-23
great-expectations < 1.0

Kilo59:


pyproject.toml
```diff
@@ -46,7 +46,7 @@ postgres = ["sqlalchemy", "psycopg2-binary"]
 databricks = ["databricks-sql-connector"]
 
 [tool.poetry.group.sql.dependencies]
-great-expectations = { version = "^0.18.19"}
+great-expectations = { version = ">=0.18.19,<1.0"}
```
anthonyburdi, line 49: Does this also need to be changed in the `[tool.poetry.dependencies]` section @Kilo59?

Kilo59: Yes! Good catch!
I thought that's what I was doing 😆

anthonyburdi (approved):
Good catch on this PR! I think we also need to do an agent release too right? To update the stable image.

Kilo59:
> I think we also need to do an agent release too right? To update the stable image.

@anthonyburdi There's no reason that we need a new `stable` image since the docker deps are always pinned to whatever is in the lockfile.
But 1 reason do a normal release now would be to prevent problems for any of our users installing us directly from `pypi`, there's nothing in place now that prevents them from installing `gx` `1.0` and breaking their agent.

pyproject.toml
```diff
@@ -23,7 +23,7 @@ include = [
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
-great-expectations = "^0.18.19"
+great-expectations = "^0.18.19,<0.19" # <0.19 instead of <v1.0 to include v1 pre-releases
```
Kilo59, line 26:
```suggestion
great-expectations = "^0.18.19,<0.19" # <0.19 instead of <1.0 to include v1 pre-releases
```

PR #411, opened 2024-08-28
Remove unneeded build-time packages gcc and cpp

wookasz:
Removes the `gcc-12`, `cpp-12`, and `cpp` from the image after build steps complete. These are unneeded and `gcc-12` is getting flagged by vulnerability scans.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,11 @@ COPY examples/agent/data data
 RUN poetry install --only-root && rm -rf POETRY_CACHE_DIR
 
 # Clean up all non-runtime linux deps
-RUN apt-get remove python3-dev gcc -y
+RUN apt-get remove -y \
+    python3-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -39,6 +39,11 @@ COPY examples/agent/data data
 RUN poetry install --only-root && rm -rf POETRY_CACHE_DIR
 
 # Clean up all non-runtime linux deps
-RUN apt-get remove python3-dev gcc -y
+RUN apt-get remove -y \
+    python3-dev \
+    gcc \
+    gcc-12 \
+    cpp-12 \
+    cpp
```
Kilo59, line 47: Aren't you removing the remove step here?

Kilo59 (approved):
Ignore my earlier comment. I can't read a diff 😆

PR #414, opened 2024-08-29
Add RunWindowCheckpointEvent

anthonyburdi:
For the upcoming Dynamic Parameters work. This is a placeholder, the action will be added in a subsequent PR.

If someone creates this event before the action is added, the agent will raise a `NoVersionImplementationError`.

### DIFF
--- a/great_expectations_cloud/agent/models.py
+++ b/great_expectations_cloud/agent/models.py
@@ -65,7 +65,6 @@ class RunScheduledCheckpointEvent(ScheduledEventBase):
     splitter_options: Optional[Dict[str, Any]] = None
 
 
-# comment to bump ci
 class RunWindowCheckpointEvent(EventBase):
     type: Literal["run_window_checkpoint.received"] = "run_window_checkpoint.received"
     datasource_names_to_asset_names: Dict[str, Set[str]]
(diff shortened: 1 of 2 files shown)


shiplet (approved):

shiplet:
🚀

great_expectations_cloud/agent/models.py
```diff
@@ -65,6 +65,13 @@ class RunScheduledCheckpointEvent(ScheduledEventBase):
     splitter_options: Optional[Dict[str, Any]] = None
 
 
+class RunWindowCheckpointEvent(EventBase):
+    type: Literal["run_window_checkpoint.received"] = "run_window_checkpoint.received"
+    datasource_names_to_asset_names: Dict[str, Set[str]]
+    checkpoint_id: uuid.UUID
+    splitter_options: Optional[Dict[str, Any]] = None
+
+
```
Kilo59, line 74: Late to the party here but why are we doing a new event here instead of a flag on the existing `RunCheckpointEvent`?
Do we expect the OSS or API calls to be significantly different?

anthonyburdi: There is going to be an extra call to determine if there are dynamic parameters, and a call to get those parameters. We are planning to do the call to determine if there are dynamic parameters in the backend to save an API call, and then use the distinct event to only do the call to get the params if there are dynamic parameters. More info in this thread: https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1268121612/Dynamic+Parameters+Technical+Design?focusedCommentId=1290502145

PR #424, opened 2024-09-09
remove the recconect logic inside of the consume function since listen is already retrying

allensallinger:


great_expectations_cloud/agent/message_service/subscriber.py
```diff
@@ -98,8 +98,6 @@ def consume(
             except KeyboardInterrupt as e:
                 self.client.stop()
                 raise KeyboardInterrupt from e
-            if self.client.should_reconnect:
-                self.client.reset()
             else:
                 break  # exit
```
Kilo59, line 103: Removing the preceeding `if` `else: break` now attaches the `elf` to the exception handling.


https://docs.python.org/3.11/tutorial/errors.html#handling-exceptions

>The [try](https://docs.python.org/3/reference/compound_stmts.html#try) … [except](https://docs.python.org/3/reference/compound_stmts.html#except) statement has an optional else clause, which, when present, must follow all except clauses. It is useful for code that must be executed if the try clause does not raise an exception

This means that if no exception is encountered we will break out of the `while` loop, which probably not what you intended.

PR #429, opened 2024-09-12
Maintenance - Remove `DataAssistant` related code and tests

Shinnnyshinshin:
Some clean up work that came up during GX Runner CI/CD epic. Removing `DataAssistant` related Actions and tests from Agent. 

Should merge #434 first


README.md
```diff
@@ -1,4 +1,4 @@
-# GX Cloud
+# GX cloud
```
rreinoldsc, line 1:
```suggestion
# GX Cloud
```

anthonyburdi (approved):
LGTM - should we also remove from the backend?

Shinnnyshinshin:
> LGTM - should we also remove from the backend?

Took a stab at it, and it's a bigger lift than just a small follow-up, with some confirmation we will need to do for the "operation drag and drop" code. Ticket was added to the backlog here. 

https://greatexpectations.atlassian.net/browse/ZELDA-983

PR #447, opened 2024-09-19
Upgrade greatexpections v1.0.0 and fix unit tests

TrangPham:
- Updates gx to v1
- fixes unit tests in tests/agent
- skips integration tests
- sets version to .a0 so it does not create another deploy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "20240918.0.a0"
+version = "20240922.0.a0"
 description = "Great Expectations Cloud"
 authors = ["The Great Expectations Team <team@greatexpectations.io>"]
 repository = "https://github.com/great-expectations/cloud"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -46,11 +48,11 @@ postgres = ["sqlalchemy", "psycopg2-binary"]
 databricks = ["databricks-sql-connector"]
 
 [tool.poetry.group.sql.dependencies]
-great-expectations = { version = ">=0.18.19"}
+great-expectations = { version = ">=0.18.19" }
```
Shinnnyshinshin, line 51: would this need to be updated too?

tests/integration/actions/conftest.py
```diff
@@ -110,31 +111,31 @@ def datasource_names_to_asset_names(datasource, data_asset):
...
-    expectation_suite = context.get_expectation_suite(expectation_suite_name=expectation_suite_name)
-    assert (
-        len(expectation_suite.expectations) == 1
-    ), "Expectation Suite was not updated in the previous method call."
-    yield expectation_suite
-    context.delete_expectation_suite(expectation_suite_name=expectation_suite_name)
-    with pytest.raises(get_missing_expectation_suite_error_type):
-        context.get_expectation_suite(expectation_suite_name=expectation_suite_name)
+# @pytest.fixture(scope="module")
+# def expectation_suite(
```
Shinnnyshinshin, line 115: Is there a reason why this is commented out and not removed?

pyproject.toml
Kilo59: yes

Shinnnyshinshin (approved):
LGTM!

PR #453, opened 2024-09-26
Revert "Upgrade greatexpections v1.0.0 and fix unit tests (#447)"

tyler-hoffman:
This reverts commit c5fe3f339a0d43a0df20429942857a67273456fe.

We're seeing the agent erroring out with `No event action map implemented for GX Core major version 1.`


billdirks (approved):
Thank you!

PR #456, opened 2024-10-01
Fix: Deploy should not consider other versions such as pre or alpha a…

TrangPham:
…s release

Tested locally:
Screenshot 2024-10-01 at 12 50 26 AM

### DIFF
--- a/.github/workflows/containerize-agent.yaml
+++ b/.github/workflows/containerize-agent.yaml
@@ -122,7 +122,7 @@ jobs:
         run: echo "IS_DEV=$(invoke is-dev-release)" >> $GITHUB_OUTPUT
 
       - name: Get is_pre_release from pyproject.toml
-        id: get_is_release
+        id: get_is_pre_release
         run: echo "IS_PRERELEASE=$(invoke is-pre-release)" >> $GITHUB_OUTPUT
 
       - name: Get is_release from pyproject.toml


tyler-hoffman (dismissed):
Thanks for doing this! Looks like there's a one character change needed to prevent false negatives, but otherwise awesome!

.github/workflows/containerize-agent.yaml
```diff
@@ -117,13 +117,17 @@ jobs:
...
         run: echo "POETRY_VERSION=$(poetry version -s)" >> $GITHUB_OUTPUT
 
+      - name: Get is_dev from pyproject.toml
+        id: get_is_dev
+        run: echo "IS_DEV=$(poetry version -s | grep -q 'dev' && echo true || echo false)" >> $GITHUB_OUTPUT
+
       - name: Get is_release from pyproject.toml
         id: get_is_release
-        run: echo "IS_RELEASE=$(poetry version -s | grep -q 'dev' && echo false || echo true)" >> $GITHUB_OUTPUT
+        run: echo "IS_RELEASE=$(poetry version -s | grep -q '[A-Za-z ]*' && echo false || echo true)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
Kilo59: You might also consider doing this with a python script/invoke task so that we are using the same method for determining versions.

See the `Version` class docs below 👇 

https://packaging.pypa.io/en/stable/version.html#packaging.version.Version

tasks.py
```diff
@@ -345,3 +345,25 @@ def release(ctx: Context, dry_run: bool = False) -> None:
...
+@invoke.task(name="is-pre-release")
+def is_pre_release(ctx: Context) -> None:
+    """Check if the current version is a pre-release."""
+    print(_get_local_version().is_prerelease and not _get_local_version().is_devrelease)
+
+
+@invoke.task(name="is-dev-release")
+def is_dev_release(ctx: Context) -> None:
+    """Check if the current version is a dev-release."""
+    print(_get_local_version().is_devrelease)
```
Kilo59, line 359:
```suggestion
    print(str(_get_local_version().is_prerelease and not _get_local_version().is_devrelease).lower())


@invoke.task(name="is-dev-release")
def is_dev_release(ctx: Context) -> None:
    """Check if the current version is a dev-release."""
    print(str(_get_local_version().is_devrelease).lower())
```
🤷 

Will print `true/false` instead of `True/False`

Kilo59 (approved):
Thanks for this!
Made 1 very minor suggestion

tasks.py
```diff
@@ -345,3 +345,25 @@ def release(ctx: Context, dry_run: bool = False) -> None:
 def prerelease(ctx: Context, dry_run: bool = False) -> None:
     """Bump project pre-release version for release to pypi and build image for dockerhub."""
     _version_bump(ctx, pre=True, standard=False, dry_run=dry_run)
+
+
+@invoke.task(name="is-pre-release")
+def is_pre_release(ctx: Context) -> None:
+    """Check if the current version is a pre-release."""
+    print(_get_local_version().is_prerelease and not _get_local_version().is_devrelease)
```
Kilo59, line 353:
```suggestion
    print(_get_local_version().is_prerelease)
```
You don't need both of these

PR #464, opened 2024-10-02
Run Window Checkpoint v0 version

anthonyburdi:
WIP - No tests are added yet. This PR likely won't get merged, instead we will merge a v1 action. Rather than write tests, if we are not using the v0 version & don't merge this PR we can just use it for internal testing of the feature.

See https://greatexpectations.atlassian.net/browse/ZELDA-959 for the Jira for the checkpoint v1 action. Once that is complete we can implement a similar workflow based on the v1 run checkpoint flow: https://docs.greatexpectations.io/docs/core/trigger_actions_based_on_results/run_a_checkpoint


anthonyburdi:
Superseded by https://github.com/great-expectations/cloud/pull/471

PR #479, opened 2024-10-10
Logging - Event obj is not used directly

Shinnnyshinshin:
Update logging in Agent so that we are long passing `Event` object directly into logs. This will make debugging much easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GX cloud
+# GX Cloud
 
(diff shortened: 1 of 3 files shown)


rreinoldsc (approved):
Always good move to allowlist the keys

shiplet (approved):

rreinoldsc (approved):

PR #480, opened 2024-10-10
Refactor to ensure more consistent usage of correlation_id

Shinnnyshinshin:
In the Agent we are using `correlation_id` and `job_id` interchangeably, which can cause confusion. This PR intends to use `correlation_id` in all locations, since it is what is used by mercury.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -282,7 +282,7 @@ class GXAgent:
             self._create_scheduled_job_and_set_started(event_context, org_id)
         else:
             self._update_status(
-                job_id=event_context.correlation_id, status=JobStarted(), org_id=org_id
+                correlation_id=event_context.correlation_id, status=JobStarted(), org_id=org_id
             )
         print(f"Starting job {event_context.event.type} ({event_context.correlation_id}) ")
         LOGGER.info(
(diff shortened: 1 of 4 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -535,7 +546,7 @@ def _set_http_session_headers(
             extra={
                 "user_agent": header_name.USER_AGENT,
                 "agent_version": agent_version,
-                "job_id": header_name.AGENT_JOB_ID,
```
Shinnnyshinshin, line 538: This is the only place that the key is being changed from `job_id` to `header_name`.  I think that is ok because

1) this is a log message, and this is the `extra` field of the message. 
2) IT is actually just a [enum](https://github.com/great-expectations/cloud/blob/bb06e6319b7300ff7bdb628caddaf500dfdd2702/great_expectations_cloud/agent/constants.py#L9) which is `Agent-Job-Id`

great_expectations_cloud/agent/agent.py
```diff
@@ -440,26 +442,35 @@ def _get_config(cls) -> GXAgentConfig:
...
             ) from validation_err
 
-    def _update_status(self, job_id: str, status: JobStatus, org_id: UUID) -> None:
+    def _update_status(self, correlation_id: str, status: JobStatus, org_id: UUID) -> None:
         """Update GX Cloud on the status of a job.
 
         Args:
-            job_id: job identifier, also known as correlation_id
-            status: pydantic model encapsulating the current status
+            correlation_id: job identifier
```
shiplet, line 449: removing the "also known as" :chefs-kiss:

PR #481, opened 2024-10-11
Error Handling after `POST` and `PATCH` requests to REST API

Shinnnyshinshin:
* When the agent runs the `_update_status()` and `_create_scheduled_job_and_get_started()` methods and hits the corresponding REST endpoints in mercury, there is no handling of any errors that may come back, which causes re-tries
* This PR intends to check the response that comes back from the BE, and raise a corresponding error. 
* [Helpful DataDog log](https://app.datadoghq.com/logs?query=service%3Agx-runner%20AND%20%28%40correlation_id%3A768f2093-12e3-48aa-a328-d68cf3b0ac9a%20OR%20%40job_id%3A768f2093-12e3-48aa-a328-d68cf3b0ac9a%20OR%20%40id%3A768f2093-12e3-48aa-a328-d68cf3b0ac9a%29%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice%2C%40correlation_id%2C%40event&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=time%2Cdesc&viz=stream&from_ts=1728436501776&to_ts=1728522901776&live=true)


great_expectations_cloud/agent/agent.py
```diff
@@ -456,11 +456,15 @@ def _update_status(self, job_id: str, status: JobStatus, org_id: UUID) -> None:
...
-            session.patch(agent_sessions_url, data=data)
+            response = session.patch(agent_sessions_url, data=data)
             LOGGER.info(
                 "Status updated",
                 extra={"job_id": job_id, "status": str(status), "organization_id": str(org_id)},
             )
+        if not response.ok:
+            raise RuntimeError(  # noqa: TRY003 # one off error
+                "Status Update action had an error while connecting to GX Cloud."
+            )
```
wookasz, line 467: Can you double-check that we return a 200 here and not a 204. Any 2XX response should be OK since we don't rely on the payload

great_expectations_cloud/agent/agent.py
```diff
@@ -501,6 +505,11 @@ def _create_scheduled_job_and_set_started(
                     "organization_id": str(org_id),
                 },
             )
+        if response.ok is not True:
+            raise GXAgentError(  # noqa: TRY003 # one off error
+                f"RunScheduledCheckpointAction encountered an error while connecting to GX Cloud."
+                f"Correlation ID of event: {event_context.correlation_id}"
+            )
```
wookasz, line 512: I believe the response to a POST is a 201

wookasz (changes requested):

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: Yup this is a `201` : https://github.com/greatexpectationslabs/mercury/blob/b78b06b47c3a3707bab35096d70922ffafd4f6f9/services/ge_cloud/mercury/handlers/rest/agent_v0.py#L183

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: This is a `204` https://github.com/greatexpectationslabs/mercury/blob/b78b06b47c3a3707bab35096d70922ffafd4f6f9/services/ge_cloud/mercury/handlers/rest/agent_v0.py#L213-L214

Shinnnyshinshin:
Thanks for the review here @wookasz, I've updated the error check to be more specific, and changed to `GXCloudError` to more-closely follow the existing pattern here: 

https://github.com/great-expectations/cloud/blob/d0ae1a3969a0916f69501b639f62a4139525f1db/great_expectations_cloud/agent/actions/list_table_names.py#L57

great_expectations_cloud/agent/agent.py
```diff
@@ -473,6 +474,11 @@ def _update_status(self, correlation_id: str, status: JobStatus, org_id: UUID) -
                     "organization_id": str(org_id),
                 },
             )
+        if not response.status_code != 204:  # noqa: PLR2004
```
Kilo59, line 477: Don't ignore these error codes.
https://docs.astral.sh/ruff/rules/magic-value-comparison/

If you don't agree with a linter rule bring it up in our COP meetings. Although I don't think it's true in this case, some of our rules could be configured better. There may even be a few that are worth disabling globally. But we don't want to get into the habit of just ignoring them.
https://requests.readthedocs.io/en/stable/api/#requests.codes
```suggestion
        if not response.status_code != requests.codes.no_content:
```

great_expectations_cloud/agent/agent.py
```diff
@@ -473,6 +474,11 @@ def _update_status(self, correlation_id: str, status: JobStatus, org_id: UUID) -
...
+            raise GXCloudError(
+                message="Status Update action had an error while connecting to GX Cloud.",
+                response=response,
+            )
```
Kilo59, line 481: This is overly specific and requires us to change the application logic should we decide to change a status code.

```suggestion
        try:
          response.raise_for_status()
        except requests.HTTPError as http_err:
          raise GXCloudError(
            message="Status Update action had an error while connecting to GX Cloud.",
            response=response,
          ) from http_err
```


A better way to handle this would be to use `response.raise_for_status()`
https://requests.readthedocs.io/en/stable/api/#requests.Response.raise_for_status

Kilo59 (changes requested):
Couple of small suggestions on improvements to handling specific response codes and errors.

tests/agent/test_agent.py
```diff
@@ -348,6 +416,82 @@ def consume(queue: str, on_message: Callable[[EventContext], None]):
...
+        # we need the main thread to remain alive until event handler has finished
+        nonlocal end_test
+        while end_test is False:
+            sleep(0)  # defer control
+
+    subscriber().consume = consume
+    agent = GXAgent()
+    create_session().__enter__().post.return_value.status_code = 404
+    agent.run()
+    create_session().__enter__().post.assert_any_call(post_url, data=payload_str)
```
Kilo59, line 492: Please use `responses` for mocking http requests.

https://github.com/getsentry/responses?tab=readme-ov-file#responses-as-a-pytest-fixture

great_expectations_cloud/agent/agent.py
```diff
@@ -513,6 +519,12 @@ def _create_scheduled_job_and_set_started(
                     "organization_id": str(org_id),
                 },
             )
+        if not response.status_code != 201:  # noqa: PLR2004
```
Kilo59, line 522: This is overly specific status code handling.
see 👉  https://github.com/great-expectations/cloud/pull/481/files#r1801193270

great_expectations_cloud/agent/agent.py
Kilo59: Please ignore my original suggestion. It fixes the linter error but misses a larger issue.

This status code handling is overly specific.
see 👉  https://github.com/great-expectations/cloud/pull/481/files#r1801193270

tests/agent/test_agent.py
```diff
@@ -279,6 +279,74 @@ def test_gx_agent_run_handles_subscriber_error_on_close(
...
+        Replicates behavior of Subscriber.consume by invoking the on_message
+        parameter with an event_context.
+        """
+        nonlocal event_context
+        on_message(event_context)
+
+        # we need the main thread to remain alive until event handler has finished
+        nonlocal end_test
+        while end_test is False:
+            sleep(0)  # defer control
```
Kilo59, line 332: `non-blocking`
I would suggest refactoring `Susbcriber.consume` (and HTTP mocking) code into separate, well-named fixtures. The inclusion of it here complicates what is otherwise a simple test

great_expectations_cloud/agent/agent.py
Shinnnyshinshin: Thank you for this @Kilo59. I think adding this method greatly simplified the testing as well..

great_expectations_cloud/agent/agent.py
```diff
@@ -581,3 +584,15 @@ def _update_headers_agent_patch(
         # TODO: this is relying on a private implementation detail
         # use a public API once it is available
         http._update_headers = _update_headers_agent_patch
+
+    def _raise_gx_cloud_err_on_http_error(self, response: requests.Response) -> None:
```
Kilo59, line 588:
```suggestion
    @classmethod
    def _raise_gx_cloud_err_on_http_error(response: requests.Response) -> None:
```
This could be a class or static method.

tests/agent/test_agent.py
```diff
@@ -605,3 +606,21 @@ def test_correlation_id_header(
...
+        agent._raise_gx_cloud_err_on_http_error(test_response)
+
+
+def test_raise_gx_cloud_err_on_http_error_success_response():
+    agent = GXAgent()
+    test_response = requests.Response()
+    # 200 - OK
+    test_response.status_code = 200
+    # no Exception raised
+    agent._raise_gx_cloud_err_on_http_error(test_response)
```
Kilo59, line 626:
```suggestion
def test_raise_gx_cloud_err_on_http_error_error_response():
    test_response = requests.Response()
    # 404 - Not Found
    test_response.status_code = 404
    with pytest.raises(gx_exception.GXCloudError):
        GXAgent._raise_gx_cloud_err_on_http_error(test_response)


def test_raise_gx_cloud_err_on_http_error_success_response():
    test_response = requests.Response()
    # 200 - OK
    test_response.status_code = 200
    # no Exception raised
    GxAgent._raise_gx_cloud_err_on_http_error(test_response)
```
You shouldn't need to create an instance of the agent to test this method.
I would strongly suggest making this a class or static method.

Kilo59 (approved):

wookasz (approved):

PR #498, opened 2024-10-21
Get view names along with tables

Kilo59:
Return views along with table names.

Given that we no longer support `QueryAsset` types in GX-Cloud, users need to be able to import views along with tables when adding assets from the UI.

https://docs.sqlalchemy.org/en/20/core/reflection.html#sqlalchemy.engine.reflection.Inspector

### DIFF
--- a/great_expectations_cloud/agent/actions/draft_datasource_config_action.py
+++ b/great_expectations_cloud/agent/actions/draft_datasource_config_action.py
@@ -69,7 +69,7 @@ class DraftDatasourceConfigAction(AgentAction[DraftDatasourceConfigEvent]):
             created_resources=[],
         )
 
-    def _get_table_names(self, datasource: Datasource, include_views: bool = False) -> list[str]:
+    def _get_table_names(self, datasource: Datasource, *, include_views: bool) -> list[str]:
         inspector: Inspector = inspect(datasource.get_engine())
         names: list[str] = inspector.get_table_names()
         if include_views:
--- a/great_expectations_cloud/agent/actions/draft_datasource_config_action.py
+++ b/great_expectations_cloud/agent/actions/draft_datasource_config_action.py
@@ -60,7 +60,7 @@ class DraftDatasourceConfigAction(AgentAction[DraftDatasourceConfigEvent]):
         datasource.test_connection(test_assets=True)  # raises `TestConnectionError` on failure
 
         if isinstance(datasource, SQLDatasource):
-            table_names = self._get_table_names(datasource=datasource)
+            table_names = self._get_table_names(datasource=datasource, include_views=True)
             self._update_table_names_list(config_id=event.config_id, table_names=table_names)
 
         return ActionResult(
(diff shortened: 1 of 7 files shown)


great_expectations_cloud/agent/actions/draft_datasource_config_action.py
```diff
@@ -69,9 +69,12 @@ def check_draft_datasource_config(
             created_resources=[],
         )
 
-    def _get_table_names(self, datasource: Datasource) -> list[str]:
+    def _get_table_names(self, datasource: Datasource, *, include_views: bool) -> list[str]:
```
wookasz, line 72: Rename to `_get_table_and_view_names` and remove the boolean flag. We don't have a need for the conditional and it makes the API less clear.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "great_expectations_cloud"
-version = "20241017.0.dev1"
+version = "20241021.0.dev0"
```
wookasz, line 3: outdated

tests/agent/actions/test_draft_datasource_config.py
```diff
@@ -170,7 +170,9 @@ def test_test_draft_datasource_config_success_sql_ds(
     assert action_result.created_resources == []
 
     # assert that the action properly calls helper methods to get table names and update the draft config
-    _get_table_names_spy.assert_called_with(datasource=datasource_cls(**datasource_config))
+    _get_table_names_spy.assert_called_with(
+        datasource=datasource_cls(**datasource_config), include_views=True
+    )
     _update_table_names_list_spy.assert_called_with(config_id=config_id, table_names=table_names)
```
wookasz, line 176: These assertions don't add any value to the test and make refactoring harder. Testing that a method calls a private helper makes no guarantee that what the helper is doing is correct.

These tests is already verifying that the correct endpoint is called with the correct tables/views and that is enough

tests/agent/actions/test_list_table_names_action.py
```diff
@@ -105,8 +105,10 @@ def test_run_list_table_names_action_returns_action_result(
     )
 
     table_names = ["table_1", "table_2", "table_3"]
+    view_names = ["view_1", "view_2"]
     inspector = mocker.Mock(spec=Inspector)
     inspector.get_table_names.return_value = table_names
+    inspector.get_view_names.return_value = view_names
```
wookasz, line 111: I would expect this test to fail since we're not getting different results. The test should guarantee that we're sending back the correct results

tests/agent/actions/test_draft_datasource_config.py
wookasz: Some tests here should guarantee that we return views as well

wookasz (changes requested):

tyler-hoffman:
Stale and not tied to a ticket; closing for now.

PR #505, opened 2024-10-22
Fix duplicate slashes in url

anthonyburdi:
We are getting errors due to duplicate slashes in urls e.g. [this error](https://app.datadoghq.com/logs?query=service%3Agx-agent-timber-pod&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAZK1uYJ70bDyFgAAAAAAAAAYAAAAAEFaSzF1WTc5QUFESDZsNUxWOXFja0FBaQAAACQAAAAAMDE5MmI1YjktZTllZC00Nzc0LTkxMDYtZTRlODJmNGRjNzhi&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1729021909080&to_ts=1729626709080&live=true).

This PR is intended to make the agent more resilient when constructing urls. In the future we can also consider using a url object instead of strings, and then construct the string when needed.


Shinnnyshinshin (approved):
looks awesome. super clear, well tested. Thanks so much for doing this @anthonyburdi

tests/agent/test_agent.py
```diff
@@ -622,3 +627,38 @@ def test_raise_gx_cloud_err_on_http_error_success_response():
...
+            "api/v1/something",
+            "http://localhost:5000/api/v1/something",
+            id="base_no_slash + path_no_slash",
+        ),
+    ],
+)
+def test_construct_url_from_base_plus_path(base: str, path: str, expected_url: str):
+    """Test that urljoin constructs all combinations of urls correctly."""
+    constructed_url = urljoin(base, path)
+    assert constructed_url == expected_url
```
rreinoldsc, line 664: no need to test `urljoin` itself

rreinoldsc (approved):
image

tests/agent/test_agent.py
anthonyburdi: Yeah you're right - I wrote this test when I had a custom fn. Will clean it up.

tests/agent/test_agent.py
```diff
@@ -285,7 +288,7 @@ def test_gx_agent_updates_cloud_on_job_status(
 ):
     correlation_id = "4ae63677-4dd5-4fb0-b511-870e7a286e77"
     url = (
-        f"{gx_agent_config.gx_cloud_base_url}/api/v1/organizations/"
+        f"http://localhost:5000/api/v1/organizations/"
```
anthonyburdi, line 291: Hardcoded because these were passing with the `//` issue.

PR #507, opened 2024-10-23
Add headers to all stores

wookasz:
- adds agent headers to remaining stores (Data Source and Asset)
- overrides GX Core user agent for all stores

Requests coming from GX Agent should have GX Agent user agent, otherwise, they're not distinguishable from notebook or script usage


Shinnnyshinshin (approved):
LGTM thank you

PR #514, opened 2024-10-25
Update get_table_names to pass in schema if snowflake datasource

TrangPham:


wookasz (approved):

Kilo59:
I don't plan to make a habit of doing these post-departure code reviews 😄, but...
@TrangPham @wookasz 

Just FYI we/ya'll also require the `schema` for all our `DatabricksSQL` datasources.
However we aren't exposing the `.schema_` as a property on the `DatabricksSQLDSN` or Datasource itself.

https://github.com/great-expectations/great_expectations/blob/abcf9f02cf756db3c5e94dea45d7cad63d16859d/great_expectations/datasource/fluent/databricks_sql_datasource.py#L109-L111

If that `my_databricks_sql.schema_` prop was added in a future OSS release you could use it here.

```python
def get_table_names(datasource: SQLDatasource) -> list[str]:
    inspector: Inspector = inspect(datasource.get_engine())
    if isinstance(datasource, SnowflakeDatasource | DatabricksSQLDatasource) and datasource.schema_:
        return inspector.get_table_names(schema=datasource.schema_)
```


Or even better you could use structural type-checking rather than constantly updating this list of datasources.
https://peps.python.org/pep-0544/#runtime-checkable-decorator-and-narrowing-types-by-isinstance

```python
from typing import Protocol, @runtime_checkable

@runtime_checkable
class HasSchema_(Protocol):
   @property
   def schema_(self) -> str | None: ...

def get_table_names(datasource: SQLDatasource) -> list[str]:
    inspector: Inspector = inspect(datasource.get_engine())
    if isinstance(datasource, HasSchema_) and datasource.schema_:
        return inspector.get_table_names(schema=datasource.schema_)
```

PR #518, opened 2024-10-29
Properly disconnect from the GX Agent if a `GXAgentUnrecoverableError` is raised

alena-hutchinson:
This is for [ZELDA-1063](https://greatexpectations.atlassian.net/browse/ZELDA-1063). 

This PR does the following: 
- Corrects the logic in the GX Agent so that we successfully disconnect if a `GXAgentUnrecoverableError` is raised
- Removes broken attempts to reconnect the client. We decided that in the interest of time, it is better to kill the connection and let Kubernetes spin a new one up. Future work can address reconnection from the client. 
- Adds structured logging to the agent and RabbitMQ client code

See [this](https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1391034381/GX+Runner+-+Long+Job+Resilience) writeup for more context about the issue. 

**Testing**
I was able to reproduce the RabbitMQ Delivery Acknowledgement Timeout locally by adding a `sleep` call to the `run_checkpoint` job. 

TODO: Verify in Linux that the command we are using to kill the agent is kosher. 

[ZELDA-1063]: https://greatexpectations.atlassian.net/browse/ZELDA-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -164,6 +164,7 @@ class GXAgent:
 
     def run(self) -> None:
         """Open a connection to GX Cloud."""
+
         LOGGER.debug("Opening connection to GX Cloud.")
         self._listen()
         # we need a heavy-handed approach here to kill the main process to prevent a scenario where the agent is
@@ -192,7 +193,6 @@ class GXAgent:
             subscriber = Subscriber(client=client)
             LOGGER.info("The GX Agent is ready.")
             # Open a connection until encountering a shutdown event
-
             subscriber.consume(
                 queue=self._config.queue,
                 on_message=self._handle_event_as_thread_enter,
@@ -326,6 +326,7 @@ class GXAgent:
             event_context: event with related properties and actions.
         """
         # warning:  this method will not be executed in the main thread
+
         org_id = self.get_organization_id(event_context)
 
         # get results or errors from the thread
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -192,6 +192,7 @@ class GXAgent:
             subscriber = Subscriber(client=client)
             LOGGER.info("The GX Agent is ready.")
             # Open a connection until encountering a shutdown event
+
             subscriber.consume(
                 queue=self._config.queue,
                 on_message=self._handle_event_as_thread_enter,
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -367,8 +367,8 @@ class GXAgent:
                 )
         else:
             status = build_failed_job_completed_status(error)
-            LOGGER.error(traceback.format_exc())
-            LOGGER.error(
+            LOGGER.info(traceback.format_exc())
+            LOGGER.info(
                 "Job completed with error",
                 extra={
                     "event_type": event_context.event.type,
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import sys
+import os
+import signal
 import traceback
 import warnings
 from collections import defaultdict
@@ -163,9 +164,12 @@ class GXAgent:
 
     def run(self) -> None:
         """Open a connection to GX Cloud."""
-
         LOGGER.debug("Opening connection to GX Cloud.")
         self._listen()
+        # we need a heavy-handed approach here to kill the main process to prevent a scenario where the agent is
+        # hanging due to a thread that has entered an unhealthy state where it will never finish, for example, in the
+        # case of a RabbitMQ delivery acknowledgement timeout
+        os.kill(os.getpid(), signal.SIGKILL)
         LOGGER.debug("The connection to GX Cloud has been closed.")
 
     # ZEL-505: A race condition can occur if two or more agents are started at the same time
@@ -198,7 +202,6 @@ class GXAgent:
             LOGGER.exception("The connection to GX Cloud has encountered an error.")
         except GXAgentUnrecoverableConnectionError:
             LOGGER.exception("The connection to GX Cloud has encountered an unrecoverable error.")
-            sys.exit(1)
         except (
             AuthenticationError,
             ProbableAuthenticationError,
@@ -322,7 +325,6 @@ class GXAgent:
             event_context: event with related properties and actions.
         """
         # warning:  this method will not be executed in the main thread
-
         org_id = self.get_organization_id(event_context)
 
         # get results or errors from the thread
@@ -364,8 +366,8 @@ class GXAgent:
                 )
         else:
             status = build_failed_job_completed_status(error)
-            LOGGER.info(traceback.format_exc())
-            LOGGER.info(
+            LOGGER.error(traceback.format_exc())
+            LOGGER.error(
                 "Job completed with error",
                 extra={
                     "event_type": event_context.event.type,
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -136,7 +136,7 @@ class GXAgent:
         agent_version: str = self.get_current_gx_agent_version()
         great_expectations_version: str = self._get_current_great_expectations_version()
         LOGGER.info(
-            "Initializing GX Agent",
+            "Initializing GX Agent.",
             extra={
                 "agent_version": agent_version,
                 "great_expectations_version": great_expectations_version,
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -136,7 +136,7 @@ class GXAgent:
         agent_version: str = self.get_current_gx_agent_version()
         great_expectations_version: str = self._get_current_great_expectations_version()
         LOGGER.info(
-            "Initializing GX Agent.",
+            "Initializing GX Agent",
             extra={
                 "agent_version": agent_version,
                 "great_expectations_version": great_expectations_version,
--- a/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
+++ b/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
@@ -238,7 +238,7 @@ class AsyncRabbitMQClient:
         # self._reconnect()
 
     def _on_connection_closed(
-        self, connection: AsyncioConnection, _unused_reason: pika.Exception
+        self, _unused_connection: AsyncioConnection, _unused_reason: pika.Exception
     ) -> None:
         """Callback invoked after the broker closes the connection"""
         # TODO: Fix reconnect logic and add a call here to reconnect if appropriate
@@ -247,7 +247,8 @@ class AsyncRabbitMQClient:
         self._is_unrecoverable = True
         # we need to call stop to break out of the run_forever call,
         # see: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_forever
-        self._connection.ioloop.stop()
+        if self._connection is not None:
+            self._connection.ioloop.stop()
 
     def _close_connection(self, reason: pika.Exception) -> None:
         """Close the connection to the broker."""
--- a/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
+++ b/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
@@ -86,11 +86,12 @@ class AsyncRabbitMQClient:
         LOGGER.debug("Shutting down the connection to RabbitMQ.")
         if not self._closing:
             self._closing = True
-
-        if self._consuming:
-            self._stop_consuming()
-        self._connection.ioloop.stop()
-        LOGGER.debug("The connection to RabbitMQ has been shut down.")
+            if self._consuming:
+                self._stop_consuming()
+                self._connection.ioloop.run_forever()
+            else:
+                self._connection.ioloop.stop()
+            LOGGER.debug("The connection to RabbitMQ has been shut down.")
 
     def reset(self) -> None:
         """Reset client to allow a restart."""
@@ -229,28 +230,24 @@ class AsyncRabbitMQClient:
         LOGGER.error(
             "Connection open failed",
             extra={
-                "reason": reason,
+                "reply_code": reason.reply_code,
+                "reply_text": reason.reply_text,
             },
         )
-        self._reconnect()
+        # TODO: Fix reconnect logic before this can be uncommented
+        # self._reconnect()
 
     def _on_connection_closed(
         self, connection: AsyncioConnection, _unused_reason: pika.Exception
     ) -> None:
         """Callback invoked after the broker closes the connection"""
+        # TODO: Fix reconnect logic and add a call here to reconnect if appropriate
         LOGGER.debug("Connection to RabbitMQ has been closed")
         self._channel = None
         self._is_unrecoverable = True
-        if self._closing:
-            connection.ioloop.stop()
-        else:
-            LOGGER.warning(
-                "Connection closed, reconnect necessary",
-                extra={
-                    "reason": reason,
-                },
-            )
-            self._reconnect()
+        # we need to call stop to break out of the run_forever call,
+        # see: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_forever
+        self._connection.ioloop.stop()
 
     def _close_connection(self, reason: pika.Exception) -> None:
         """Close the connection to the broker."""
--- a/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
+++ b/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
@@ -226,14 +226,13 @@ class AsyncRabbitMQClient:
         self, _unused_connection: AsyncioConnection, reason: pika.Exception
     ) -> None:
         """Callback invoked when there is an error while opening connection."""
-        self._reconnect()
         LOGGER.error(
             "Connection open failed",
             extra={
-                "reply_code": reason.reply_code,
-                "reply_text": reason.reply_text,
+                "reason": reason,
             },
         )
+        self._reconnect()
 
     def _on_connection_closed(
         self, connection: AsyncioConnection, _unused_reason: pika.Exception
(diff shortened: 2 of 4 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -159,10 +160,13 @@ def __init__(self: Self):
...
-
-        print("Opening connection to GX Cloud.")
+        LOGGER.info("Opening connection to GX Cloud.")
         self._listen()
-        print("The connection to GX Cloud has been closed.")
+        # we need a heavy-handed approach here to kill the main process to prevent a scenario where the agent is
+        # hanging due to a thread that has entered an unhealthy state where it will never finish, for example, in the
+        # case of a RabbitMQ delivery acknowledgement timeout
+        os.kill(os.getpid(), signal.SIGKILL)
+        LOGGER.info("The connection to GX Cloud has been closed.")
```
Shinnnyshinshin, line 169: 🔪  `os.kill()`

great_expectations_cloud/agent/agent.py
```diff
@@ -134,17 +135,17 @@ class GXAgent:
     def __init__(self: Self):
         agent_version: str = self.get_current_gx_agent_version()
         great_expectations_version: str = self._get_current_great_expectations_version()
-        print(f"GX Agent version: {agent_version}")
-        print(f"Great Expectations version: {great_expectations_version}")
-        print("Initializing the GX Agent.")
+        LOGGER.info(f"GX Agent version: {agent_version}")
```
wookasz, line 138: nit, can we put dynamic values in `extra`?

great_expectations_cloud/agent/agent.py
```diff
@@ -361,8 +364,8 @@ def _handle_event_as_thread_exit(
                 )
         else:
             status = build_failed_job_completed_status(error)
-            LOGGER.info(traceback.format_exc())
-            LOGGER.info(
+            LOGGER.error(traceback.format_exc())
+            LOGGER.error(
```
wookasz, line 368: I think this should stay as `info`. While the job errored, that's not necessarily (and not likely) an error in the agent. More likely it's something like incorrect credentials, attempting to test a non-existent table or column, etc.

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -79,15 +83,19 @@ def stop(self) -> None:
...
+        LOGGER.info("Stopping")
         if not self._closing:
             self._closing = True
-
-        if self._consuming:
-            self._stop_consuming()
-        self._connection.ioloop.stop()
+            if self._consuming:
+                self._stop_consuming()
+                self._connection.ioloop.run_forever()
```
anthonyburdi, line 91: Wondering why we are calling run_forever in the stop()?

great_expectations_cloud/agent/agent.py
wookasz: Actually do we need 3 messages here? Could we do just one `Initializing the GX Agent` message with `agent_version` and `gx_core_version` in `extra`?

great_expectations_cloud/agent/agent.py
```diff
@@ -134,17 +135,17 @@ class GXAgent:
...
         great_expectations_version: str = self._get_current_great_expectations_version()
-        print(f"GX Agent version: {agent_version}")
-        print(f"Great Expectations version: {great_expectations_version}")
-        print("Initializing the GX Agent.")
+        LOGGER.info(f"GX Agent version: {agent_version}")
+        LOGGER.info(f"Great Expectations version: {great_expectations_version}")
+        LOGGER.info("Initializing the GX Agent.")
         self._config = self._get_config()
-        print("Loading a DataContext - this might take a moment.")
+        LOGGER.info("Loading a DataContext - this might take a moment.")
```
wookasz, line 142: I think generally the types of messages that are intended to determine when a section of code starts and ends should be at `debug` level. They're generally only useful when debugging

great_expectations_cloud/agent/agent.py
```diff
@@ -159,10 +160,14 @@ def __init__(self: Self):
 
     def run(self) -> None:
         """Open a connection to GX Cloud."""
+        LOGGER.info("Opening connection to GX Cloud.")
```
wookasz, line 163: Consider `debug` for this and the connection closed message

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -79,15 +83,19 @@ def stop(self) -> None:
         """Close the connection to RabbitMQ."""
         if self._connection is None:
             return
+        LOGGER.info("Stopping")
```
wookasz, line 86: `debug` and maybe elaborate more on what is stopping

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -79,15 +83,19 @@ def stop(self) -> None:
...
-
-        if self._consuming:
-            self._stop_consuming()
-        self._connection.ioloop.stop()
+            if self._consuming:
+                self._stop_consuming()
+                self._connection.ioloop.run_forever()
+            else:
+                self._connection.ioloop.stop()
+        LOGGER.info("Stopped")
```
wookasz, line 94: `debug` and maybe elaborate more on what stopped

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -79,15 +83,19 @@ def stop(self) -> None:
...
 
     def reset(self) -> None:
         """Reset client to allow a restart."""
+        LOGGER.info("Resetting client")
```
wookasz, line 98: this _feels_ `debug`

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
 
     def _start_consuming(self, queue: str, on_message: OnMessageFn, channel: Channel) -> None:
         """Consume from a channel with the on_message callback."""
+        LOGGER.info("Issuing consumer-related RPC commands")
+        LOGGER.info("Adding consumer cancellation callback")
```
wookasz, line 185: Do we need two messages here?

These seem to be `debug`

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
+                "Consumer was cancelled remotely, shutting down",
+                extra={
+                    "method_frame": method_frame,
+                },
+            )
             self._channel.close()
 
     def _reconnect(self) -> None:
         """Prepare the client to reconnect."""
+        LOGGER.info("Preparing client to reconnect")
```
wookasz, line 202: Maybe `debug`

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
         self.should_reconnect = True
         self.stop()
 
     def _stop_consuming(self) -> None:
         """Cancel the channel, if it exists."""
         if self._channel is not None:
+            LOGGER.info("Sending a Basic.Cancel RPC command to RabbitMQ")
```
wookasz, line 209: This feels like a `debug` message

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
             self._channel.basic_cancel(self._consumer_tag, callback=self._on_cancel_ok)
 
-    def _on_cancel_ok(self, method_frame: Basic.CancelOk) -> None:
+    def _on_cancel_ok(self, _unused_frame: Basic.CancelOk) -> None:
         """Callback invoked after broker confirms cancel."""
         self._consuming = False
         if self._channel is not None:
+            LOGGER.info("RabbitMQ acknowledged the cancellation of the consumer")
```
wookasz, line 216: `debug`?

anthonyburdi (commented):
Just a question

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
             self._channel.close()
 
     def _on_connection_open(
         self, connection: AsyncioConnection, queue: str, on_message: OnMessageFn
     ) -> None:
         """Callback invoked after the broker opens the connection."""
+        LOGGER.info("Connection opened")
```
wookasz, line 223: `debug`?

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
+        # TODO: Fix reconnect logic before this can be uncommented
+        # self._reconnect()
 
-    def _on_connection_closed(self, connection: AsyncioConnection, reason: str) -> None:
+    def _on_connection_closed(
+        self, _unused_connection: AsyncioConnection, reason: Exception
+    ) -> None:
         """Callback invoked after the broker closes the connection"""
+        # TODO: Fix reconnect logic and add a call here to reconnect if appropriate
+        LOGGER.info("Connection closed")
```
wookasz, line 246: `debug`? we have other messages informing us when shutdown happens

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
+        # we need to call stop to break out of the run_forever call,
+        # see: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_forever
+        self._connection.ioloop.stop()
 
-    def _close_connection(self) -> None:
+    def _close_connection(self, reason: Exception) -> None:
         """Close the connection to the broker."""
         self._consuming = False
         if self._connection is None or self._connection.is_closing or self._connection.is_closed:
+            LOGGER.info("Connection closing or is already closed")
```
wookasz, line 257: `debug`?

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
             pass
         else:
-            self._connection.close()
+            LOGGER.info("Closing connection")
```
wookasz, line 260: `debug`?

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -173,67 +181,103 @@ def _callback_handler(
...
+            self._connection.close(reply_code=reason.reply_code, reply_text=reason.reply_text)
 
     def _on_channel_open(self, channel: Channel, queue: str, on_message: OnMessageFn) -> None:
         """Callback invoked after the broker opens the channel."""
+        LOGGER.info("Channel opened")
```
wookasz, line 265: `debug`?

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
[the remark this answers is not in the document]
wookasz: Nice find, that is really interesting

great_expectations_cloud/agent/agent.py
```diff
@@ -159,10 +160,14 @@ def __init__(self: Self):
...
 
-        print("Opening connection to GX Cloud.")
         self._listen()
-        print("The connection to GX Cloud has been closed.")
+        # we need a heavy-handed approach here to kill the main process to prevent a scenario where the agent is
+        # hanging due to a thread that has entered an unhealthy state where it will never finish, for example, in the
+        # case of a RabbitMQ delivery acknowledgement timeout
+        os.kill(os.getpid(), signal.SIGKILL)
```
wookasz, line 169: This doesn't feel right. Why/how is the thread getting into a state where it can never finish? Can we do something on that thread to make sure it's healthy? Which thread is getting stuck?

wookasz:
@alena-hutchinson would you be willing to make the exit changes a separate PR form the logging changes?

It's easier to roll back if we need to and also easier to look at/reason about the critical changes on their own.

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
alena-hutchinson: https://github.com/great-expectations/cloud/pull/518#discussion_r1823407227

great_expectations_cloud/agent/agent.py
alena-hutchinson: Yeah, it does feel gross... but there was a lot of back and forth trying to figure out how to do this between myself, @anthonyburdi and @billdirks. 

For context: https://greatexpectationslabs.slack.com/archives/C05R8RE43CL/p1730269919669839?thread_ts=1730242168.373279&cid=C05R8RE43CL

Happy to discuss more, but what was happening is that the `ThreadPoolExecutor` we define [here](https://github.com/great-expectations/cloud/blob/main/great_expectations_cloud/agent/agent.py#L155) was stuck waiting on its worker thread when a DAT error would get thrown. 

If you have any other ideas about how we can shut down, super interested!

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -16,6 +17,9 @@
     from pika.channel import Channel
     from pika.spec import Basic, BasicProperties
 
+LOGGER: Final[logging.Logger] = logging.getLogger(__name__)
+LOGGER.setLevel(logging.DEBUG)
```
wookasz, line 21: Not for this PR, but at some point we'll likely want to control these with an environment variable so that users will be able to change the logging level for their deployments

alena-hutchinson:
> @alena-hutchinson would you be willing to make the exit changes a separate PR form the logging changes?
> 
> It's easier to roll back if we need to and also easier to look at/reason about the critical changes on their own.

Pulled the logging changes out: https://github.com/great-expectations/cloud/pull/522

great_expectations_cloud/agent/agent.py
billdirks: I don't think the worker thread is actually in an unhealthy state. I think what is going on is:
1. This thread is the one connected to rabbit (the `_listen` call above). Our rabbit connection goes into a bad state (eg rabbit dies for an upgrade), so we exit the _listen call.
2. The other thread is the worker thread. It is happily working but may be doing a task that takes a long time. If this task were to complete we have no way of acknowledging it to rabbit since that connection is broken.
3. Because of this we want to exit the process. Exiting this thread (eg `sys.exit()`) doesn't exit the whole process. This is because we are using a `ThreadPoolExecutor`. See the [docs here](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor):
>  All threads enqueued to ThreadPoolExecutor will be joined before the interpreter can exit. Note that the exit handler which does this is executed before any exit handlers added using atexit. This means exceptions in the main thread must be caught and handled in order to signal threads to exit gracefully. For this reason, it is recommended that ThreadPoolExecutor not be used for long-running tasks.
4. Since we have no API way of forcing an exit we `kill`.

I didn't understand all these details when we talked yesterday.

Probably the easiest better solution be to avoid the `ThreadPoolExecutor` altogether. Our pool is of size 1 here, we could use the threading API directly. We could make the worker thread a `daemon` thread so it wouldn't keep the process alive if the main thread exited. Even if we spawned a new thread for each task continued running only 1 task at a time, our performance would probably be fine here since that should be fast with respect to job length. 

A harder, but doable, solution would be to implement our own `Executor`, as the docs warn this one isn't meant for long lived tasks. Forcing the call to join to complete before exiting seems like a detractor to this implementation.

wookasz: @billdirks thanks for the clarification.

To keep scope of this fix low, let's keep the `kill` and put more thought into a longer-term solution.

I think we should also consider:
- whether in fact we should be timing out/killing jobs like this. If we find that validation runs are taking over 2 hours, we should figure out how to make them faster or how to restrict the amount of work per single task to make sure they finish
- whether we're certain that we want to do the ACK after the job completes instead of moving it before

great_expectations_cloud/agent/agent.py
```diff
@@ -159,10 +164,14 @@ def __init__(self: Self):
...
+        LOGGER.debug("Opening connection to GX Cloud.")
 
-        print("Opening connection to GX Cloud.")
         self._listen()
-        print("The connection to GX Cloud has been closed.")
+        # we need a heavy-handed approach here to kill the main process to prevent a scenario where the agent is
+        # hanging due to a thread that has entered an unhealthy state where it will never finish, for example, in the
+        # case of a RabbitMQ delivery acknowledgement timeout
+        os.kill(os.getpid(), signal.SIGKILL)
+        LOGGER.debug("The connection to GX Cloud has been closed.")
```
wookasz, line 174: This should get moved before the `kill`

wookasz (approved):

great_expectations_cloud/agent/agent.py
billdirks: One other thought (I was talking to @Shinnnyshinshin about this), is maybe we can manage the connection to rabbitmq better. However, we're trying to reconnect now isn't working. I haven't looked at how that is implemented (are we trying to use the same connection object? What are it's semantics?). Maybe we need to reinitiate the connection de novo?

alena-hutchinson: We've moved conversation to [this](https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1431764993/GX+Agent+Runner+Resilience) doc. It goes into more detail about the issues with reconnecting.

alena-hutchinson:
Closing this because the Agent does shut down, just not until the worker thread finishes. See future work to improve resilience to long-running jobs [here](https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1431764993/GX+Agent+Runner+Resilience).

PR #520, opened 2024-10-30
Add job timeouts

anthonyburdi:


anthonyburdi:
See the “Adding a timeout within job thread“ section of the doc for an explanation of the issues with this approach: [GX Agent/Runner Resilience | Adding a timeout within job thread](https://greatexpectations.atlassian.net/wiki/spaces/SUP/pages/1431764993/GX+Agent+Runner+Resilience#Adding-a-timeout-within-job-thread). Basically it won't work as implemented in this PR, we would need additional (possibly large) changes.

PR #530, opened 2024-11-08
remove scheduled window event; default to checking for windows in all…

shiplet:
… scheduled events

Having a segmentation between windowed & non-windowed scheduled events creates an issue where the first expectation type saved to eventbridge becomes the norm for all subsequent runs of the checkpoint, regardless of whether there's windowed or fixed expectations.

This leads to cases where a fixed expectation as the first in a suite saves a fixed action to the scheduler, and precludes the windowed code-paths for any windowed expectations from ever running.

To fix this, we're doing away with the dual concept of scheduled/windowed-scheduled events, and treating every scheduled event as one with possible windows in place.


tests/agent/actions/checkpoints/test_run_checkpoint_action.py
```diff
@@ -79,89 +102,180 @@ def cloud_base_url() -> str:
...
     ],
 )
+@responses.activate
 def test_run_checkpoint_action_with_and_without_splitter_options_returns_action_result(
-    mock_context, action_class, event, splitter_options, batch_request, cloud_base_url
+    mock_context,
+    mocker: MockerFixture,
+    action_class,
+    event,
+    mock_patch,
```
Shinnnyshinshin, line 118: @shiplet could you tell me a little bit more about how `mock_patch` works here?

shiplet: Oh yeah, so in the cases where `mock_patch` exists in either the `run_checkpoint_action_class_and_event` or `run_scheduled_checkopint_action_class_and_event` tuples, it's a string representing the python path of the function that we're mocking/spying on, so we can track what it was called with. 

The only downside to this pattern is I've gotta include it in all of the function signatures of the tests that use those tuples, or else the pytest runner throws an error saying it can't find the value.

shiplet:
@Shinnnyshinshin Yes, this would require a migration to switch all schedules to a `run_scheduled_checkpoint.received` type. Good catch.

PR #537, opened 2024-11-14
Try catch around update status

TrangPham:
Adds a try-catch around _update_status as that could raise an error which would then trigger the done callback in an infinite loop.

nacks the message without redelivery as well.


anthonyburdi (changes requested):
See comment about adding a return, needs to happen before merge.

anthonyburdi (approved):
Thanks for adding tests! LGTM!

PR #540, opened 2024-11-15
Log only if certain type of error is found

anthonyburdi:
Some exception types do not have a `reply_code` or `reply_text`. Before logging, we check the exception type.

### DIFF
--- a/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
+++ b/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
@@ -255,8 +255,8 @@ class AsyncRabbitMQClient:
                 reply_code = reason.reply_code
                 reply_text = reason.reply_text
             else:
-                reply_code = 0
-                reply_text = "Unknown error"
+                reply_code = 999  # arbitrary value, not in the list of AMQP reply codes: https://www.rabbitmq.com/amqp-0-9-1-reference#constants
+                reply_text = str(reason)
             self._connection.close(reply_code=reply_code, reply_text=reply_text)
 
     def _log_pika_exception(
--- a/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
+++ b/great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Callable, Final, Protocol
 
 import pika
 from pika.adapters.asyncio_connection import AsyncioConnection
+from pika.exceptions import ChannelClosed, ConnectionClosed
 
 from great_expectations_cloud.agent.exceptions import GXAgentUnrecoverableConnectionError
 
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
```diff
@@ -255,7 +250,32 @@ def _close_connection(self, reason: pika.Exception) -> None:
...
             LOGGER.debug("Closing connection to RabbitMQ")
-            self._connection.close(reply_code=reason.reply_code, reply_text=reason.reply_text)
+
+            if isinstance(reason, (ConnectionClosed, ChannelClosed)):
+                reply_code = reason.reply_code
+                reply_text = reason.reply_text
+            else:
+                reply_code = 0
+                reply_text = "Unknown error"
+            self._connection.close(reply_code=reply_code, reply_text=reply_text)
```
anthonyburdi, line 260: Not using `_log_pika_exception()` here since we are using the `reply_code` and `reply_text` in `close()`, not logging it.

shiplet (approved):
lgtm!

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
alena-hutchinson: Did we test that this `close` call accepts an input of `0`?

Shinnnyshinshin (approved):

great_expectations_cloud/agent/message_service/asyncio_rabbit_mq_client.py
anthonyburdi: I think this is more of an escape hatch if we encounter an unexpected exception. Maybe better would be to make the `reply_text` a string repr of the reason. As for `0`, good point - I think that is the success code. Trying to find the link to the list of codes.

alena-hutchinson: The [default](https://pika.readthedocs.io/en/stable/modules/connection.html#pika.connection.Connection.close) looks like a 200, that's why I was concerned about the 0

anthonyburdi: Yes - good point about the 200. I was mistakenly looking at the channel.close() not connection.close().

anthonyburdi: Looks like this may be the list of codes: https://www.rabbitmq.com/amqp-0-9-1-reference#constants. I think we should use something not on this list since there is no "unknown" error.

anthonyburdi: Changed in https://github.com/great-expectations/cloud/pull/540/commits/6c132e1b1616ca230737bad63ab5b3c7821a3f97

PR #551, opened 2024-11-21
`GenerateSchemaChangeExpectationAction`

Shinnnyshinshin:


great_expectations_cloud/agent/actions/generate_schema_change_expectations_action.py
```diff
@@ -52,30 +58,87 @@ def __init__(  # noqa: PLR0913  # Refactor opportunity
...
+        return metric_run, metric_run_id
+
+    def _add_schema_change_expectation(
+        self, metric_run: MetricRun, expectation_suite_name: str
+    ) -> gx_expectations.Expectation:
+        try:
+            expectation_suite = self._context.suites.get(name=expectation_suite_name)
+        except DataContextError as e:
+            raise RuntimeError(  # noqa: TRY003 # want to keep this informative for now
+                f"Expectation Suite with name {expectation_suite_name} was not found."
```
shiplet, line 122: In the event we don't find the suite, is there anything in place that creates one?

shiplet (approved):
Only one question, otherwise lgtm!

great_expectations_cloud/agent/actions/generate_schema_change_expectations_action.py
Shinnnyshinshin: Hey good question. I took that out because of the SUJE work, which will create a default GX-managed ExpectationSuite along with the asset.  

The format will look like this. 
```
<data_asset_name> <random 6 char hash> - GX-Managed Checkpoint"
"<data_asset_name> <random 6 char hash> - GX-Managed Expectation Suite"
```

So the appropriate response here would be to report the error back (as part of the ZELDA-1153 and ZELDA-1127)

shiplet: Awesome, thanks for clarifying!

great_expectations_cloud/agent/actions/generate_schema_change_expectations_action.py
```diff
@@ -52,30 +58,87 @@ def __init__(  # noqa: PLR0913  # Refactor opportunity
...
+        self, event: GenerateSchemaChangeExpectationsEvent, asset_name: str
+    ) -> DataAsset:
+        try:
+            datasource = self._context.data_sources.get(event.datasource_name)
+            data_asset = datasource.get_asset(asset_name)
+            data_asset.test_connection()  # raises `TestConnectionError` on failure
+
+        except Exception as e:
+            # TODO - see if this can be made more specific
+            raise RuntimeError(f"Failed to retrieve asset: {e}") from e  # noqa: TRY003 # want to keep this informative for now
```
anthonyburdi, line 97: What do you think about removing this try/except here? The try/except in the run method will catch any error raised here, and will report the asset name and error (line 79).
Eventually in ZELDA-1153 we can collect any raised exceptions and re-raise in the except block of the `run` method so that the event/action handling thread can catch it and report it back to the backend.

great_expectations_cloud/agent/actions/generate_schema_change_expectations_action.py
```diff
@@ -52,30 +58,87 @@ def __init__(  # noqa: PLR0913  # Refactor opportunity
...
+
+        return data_asset
+
+    def _get_metrics(self, data_asset: DataAsset) -> tuple[MetricRun, UUID]:
+        batch_request = data_asset.build_batch_request()
+        metric_run = self._batch_inspector.compute_metric_list_run(
+            data_asset_id=data_asset.id,
+            batch_request=batch_request,
+            metric_list=[MetricTypes.TABLE_COLUMNS, MetricTypes.TABLE_COLUMN_TYPES],
+        )
```
anthonyburdi, line 107: This isn't the PR to do it in, but soon we should be using the newer v1 style of getting a batch.

great_expectations_cloud/agent/actions/generate_schema_change_expectations_action.py
anthonyburdi: Actually, I see how you are raising other errors in the methods that get executed in the `run` method to be descriptive about which part of the process failed. That makes sense. Maybe we can consider more custom error types rather than `RuntimeError` e.g. `AssetRetrievalError` but I'm ok with this for now.

great_expectations_cloud/agent/models.py
```diff
@@ -109,6 +109,8 @@ class GenerateSchemaChangeExpectationsEvent(EventBase):
     )
     datasource_name: str
     data_assets: Sequence[str]
+    # TODO - Enable this in ZELDA-1154
+    # data_asset_to_expectation_suite_name: Dict[str, str]
```
anthonyburdi, line 113: TY for the helpful comment here.

anthonyburdi (approved):
Looks good! Just a couple comments but nothing blocking.

PR #555, opened 2024-11-21
Capture exception via logs

anthonyburdi:
Previously, the exception raised when either updating the job status or creating a job in the backend for scheduled jobs would show as a multi-line exception in datadog. Since these errors are raised during the processing of an action, they should be logged rather than raised. Other errors raised during the processing of an action are already successfully caught and logged.

This PR uses our existing json logging framework to log the error as json.

Example new log:
```
{"args": "()", "levelname": "'ERROR'", "filename": "'agent.py'", "module": "'agent'", "exc_info": "None", "stack_info": "None", "funcName": "'_log_http_error'", "created": "1732221296.059027", "relativeCreated": "11169.095039367676", "response": "<Response [204]>", "error_message": "'Status Update action had an error while connecting to GX Cloud.'", "event": "'Traceback (most recent call last):\\n  File \"/path_to_code/cloud/great_expectations_cloud/agent/agent.py\", line 669, in _log_http_error\\n    raise requests.HTTPError\\nrequests.exceptions.HTTPError\\n'", "level": "'ERROR'", "logger": "'great_expectations_cloud.agent.agent'", "timestamp": "'2024-11-21T20:34:56.059027+00:00'"}
```

The previous log stretched across many datadog events:
image

PR #556, opened 2024-11-21
WIP enhance agent logs

anthonyburdi:
Setting this aside for now, this was an attempt to add structlog type logging to the agent, similar to our other repos. I erroneously thought our json logging was not working at all, but it turned out to only not be working for a single exception. That was addressed in: https://github.com/great-expectations/cloud/pull/555.


anthonyburdi:
Closing for now.

PR #558, opened 2024-11-22
Disable analytics in OSS code

TrangPham:
See for context: https://github.com/greatexpectationslabs/infrastructure/pull/311/files#r1854502407

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,6 @@ RUN apt-get remove -y \
     cpp
 
 # Disable analytics in OSS
-RUN export GX_ANALYTICS_ENABLED=false
+ENV GX_ANALYTICS_ENABLED=false
 
 ENTRYPOINT ["poetry", "run", "gx-agent"]


Dockerfile
```diff
@@ -46,4 +46,7 @@ RUN apt-get remove -y \
     cpp-12 \
     cpp
 
+# Disable analytics in OSS
+RUN export GX_ANALYTICS_ENABLED=false
```
anthonyburdi, line 50: Any reason to not use the `ENV` instruction? https://docs.docker.com/reference/dockerfile/#env

anthonyburdi (approved):
Just a question, but if this works no need to block.

Dockerfile
billdirks, line 50: nit: Can you capitalize `False` to match our docs?

billdirks (approved):
LGTM, thanks!

Dockerfile
billdirks: Oh good catch, you should use ENV. I think RUN just sets it for this session when building the image but ENV will set it in the resultant image. At least according to this: https://stackoverflow.com/questions/33379393/docker-env-vs-run-export

PR #586, opened 2024-12-09
Reenable integration test

anthonyburdi:
This test was failing because it could not connect to the backend at port 7000. Once we realized that both v0 and v1 are being served on port 5000 via this PR: https://github.com/great-expectations/cloud/commit/b84f86b7405b0b86bf63e2ea7986cb48eb674485, it meant that this test should also use port 5000. It also uses the common fixture instead of a test specific fixture.

### DIFF
--- a/tests/integration/actions/test_generate_schema_change_expectations.py
+++ b/tests/integration/actions/test_generate_schema_change_expectations.py
@@ -55,7 +55,7 @@ def seed_and_cleanup_test_data(context: CloudDataContext):
     )
 
     # Mark the validation as gx_managed
-    engine = sa.create_engine("postgresql://postgres:postgres@db:5432/mercury")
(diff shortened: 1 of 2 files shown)


tests/integration/actions/test_generate_schema_change_expectations.py
```diff
@@ -94,24 +94,16 @@ def org_id_env_var_local():
...
 
-@pytest.mark.skip(
-    "This test fails due to a connection error when attempting to call the backend. It is a problem with the test setup, not the code which was demoed, and this test passes locally."
-)
 def test_running_schema_change_expectation_action(
     context: CloudDataContext,
     user_api_token_headers_org_admin_sc_org,
     org_id_env_var_local: str,
-    cloud_base_url_local: str,
+    cloud_base_url: str,
```
anthonyburdi, line 106: Using the common fixture instead

PR #583, opened 2024-12-06
Create expectation with autogenerated flag set to true

anthonyburdi:
## What

Create an expectation with the autogenerated flag set to true. Since the autogenerated field is not a GX Core concept, we add it to the expectation config before sending to mercury.

Note that this PR skips the existing integration test. We will have a followup PR to re-enable. For now, here is a demo of this functionality working e2e with both the agent and the backend: https://gxlabs.zoom.us/clips/share/A2F3MRZZZU5JUEpSWFRKLXN2X0NtMExrazZBAQ

- App changes: https://github.com/greatexpectationslabs/mercury/pull/4114
- REST API changes: https://github.com/greatexpectationslabs/mercury/pull/4118
- Agent changes: This PR: https://github.com/great-expectations/cloud/pull/583

## Why

We need to determine whether an expectation is autogenerated or not so that we can handle differently in the UI.

PR #587, opened 2024-12-10
Remove extra `data_asset_to_expectation_suite_name` mapping from Event

Shinnnyshinshin:
Removing `data_asset_to_expectation_suite_name` mapping which is no longer needed since: 

1. after the REST API changes [here](https://github.com/greatexpectationslabs/mercury/pull/4118)
2. and cloud changes [here](https://github.com/great-expectations/cloud/pull/583)

We no longer need to provide a mapping from the Data Asset to Expectation Suite because we cause the REST endpoint directly, and it has access to the Suite.


shiplet (approved):
🚀

anthonyburdi (approved):
🎉

alena-hutchinson (approved):

PR #577, opened 2024-12-04
Schema Change - Update Event Name

Shinnnyshinshin:
* Rename schema change actions to be a bit more generic - `GenerateDataQualityCheckExpectationsEvent` and `GenerateDataQualityCheckExpectationsAction`

PR #589, opened 2024-12-11
Schema Change - Update Event Name (v2)

Shinnnyshinshin:
* Re-opened from #577 to overcome merge conflicts

* Renaming Schema Change actions and events to be more general
  * Rename `GenerateSchemaChangeExpectationsAction` to `GenerateDataQualityCheckExpectationsAction` 
  * Rename `generate_schema_change_expectations_request` to `generate_data_quality_check_expectations_request`


shiplet (approved):

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -129,7 +131,7 @@ def _get_metrics(self, data_asset: DataAsset) -> tuple[MetricRun, UUID]:
 
         return metric_run, metric_run_id
 
-    def _add_schema_change_expectation(self, metric_run: MetricRun, asset_id: UUID) -> UUID:
+    def _add_data_quality_check_expectation(self, metric_run: MetricRun, asset_id: UUID) -> UUID:
```
anthonyburdi, line 134: Nit: I would leave this one as schema change, then we can add another method for other expectation types. That way it is easy to read the `run` method and know what we are adding. But we can change when we add more expectations here, not a rush.

Shinnnyshinshin: Sure that's a really good catch. Since it's just an internal method, it should be easy to fix and get out with the next release

PR #591, opened 2024-12-12
Follow up to python 3.8 upgrade.  `ruff` and `mypy` updates

Shinnnyshinshin:
* Updating `black`, `mypy` and `ruff` to use python 3.8. 
* Moving from `Dict` -> to `dict` typing and `Set` -> `set` typing as a result. 
* Some automated formatting changes too.


roblim (approved):
Nice! LGTM!

PR #594, opened 2024-12-13
Rename `_add_data_quality_check_expectation` to be `_add_schema_change_expectation`, which is more accurate

Shinnnyshinshin:
* renaming an internal method to `_add_schema_change_expectation`  which will allow more autogenerated Expectations

### DIFF
--- a/tests/integration/actions/test_draft_datasource_config_action.py
+++ b/tests/integration/actions/test_draft_datasource_config_action.py
@@ -66,6 +66,7 @@ def test_running_draft_datasource_config_action(
         "system_users",
         "checkpoints",
         "organizations",
+        "organizations_auth0_orgs",
         "metric_runs",
         "users",
         "auth0_users",
--- a/tests/integration/actions/test_list_table_names.py
+++ b/tests/integration/actions/test_list_table_names.py
@@ -49,6 +49,7 @@ def test_running_list_table_names_action(
         "datasources",
         "expectation_suites",
         "organizations",
+        "organizations_auth0_orgs",
         "api_tokens",
         "users",
         "agent_job_created_resources",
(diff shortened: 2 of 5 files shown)


tests/integration/actions/test_draft_datasource_config_action.py
```diff
@@ -66,6 +66,7 @@ def test_running_draft_datasource_config_action(
         "system_users",
         "checkpoints",
         "organizations",
+        "organizations_auth0_orgs",
```
anthonyburdi, line 69: Any idea why this was added?

tests/integration/actions/test_list_table_names.py
```diff
@@ -49,6 +49,7 @@ def test_running_list_table_names_action(
         "datasources",
         "expectation_suites",
         "organizations",
+        "organizations_auth0_orgs",
```
anthonyburdi, line 52: Any idea why this was added?

anthonyburdi (changes requested):
LGTM - just not sure why the test table names were added.

tests/integration/actions/test_draft_datasource_config_action.py
Shinnnyshinshin: sorry I should have added an explanation here: 

1. Lakitu created a auth0 table in mercury as part of their work here: https://github.com/greatexpectationslabs/mercury/pull/4141, which is being now being added to the DB that is used for integration tests for `cloud` and `runner`.
2. Runner has the equivalent change in the corresponding test [here](https://github.com/greatexpectationslabs/gx-runner/blob/3696e3b939b4f194fe887a173347d7e475c29de9/tests/integration/actions/test_draft_datasource_config_action.py#L68)

anthonyburdi: All good - thanks for explaining here!

anthonyburdi (approved):

PR #613, opened 2025-01-08
Add temporary logging to determine if jobs are being re-queued

alena-hutchinson:
Part of work to investigate missing/inconsistent schedules ([ZELDA-1194](https://greatexpectations.atlassian.net/jira/software/c/projects/ZELDA/boards/85?assignee=712020%3Aa6524cc2-4928-4a00-924a-33f2eccde893&selectedIssue=ZELDA-1194)).
See Slack thread [here](https://greatexpectationslabs.slack.com/archives/C05R8RE43CL/p1736356433906939) for more context.  

[ZELDA-1194]: https://greatexpectations.atlassian.net/browse/ZELDA-1194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -234,7 +234,15 @@ class GXAgent:
             event_context: An Event with related properties and actions.
         """
         if self._reject_correlation_id(event_context.correlation_id) is True:
-            # this event has been redelivered too many times - remove it from circulation
+            logging.info(
+                "This event has exceeded its redelivery limit, removing from circulation",
+                extra={
+                    "event_type": event_context.event.type,
+                    "correlation_id": event_context.correlation_id,
+                    "organization_id": event_context.organization.id,
+                    "schedule_id": event_context.event.schedule_id,
+                },
+            )
             event_context.processed_with_failures()
             return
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -238,6 +238,15 @@ def _handle_event_as_thread_enter(self, event_context: EventContext) -> None:
             event_context.processed_with_failures()
             return
         elif self._can_accept_new_task() is not True:
+            logging.info(
```
shiplet, line 241: Do we also want to add a log to the early return a few lines above this, where we boot the event from circulation? I expect we can infer from the number of `cannot accept new task` logs that the event got booted, but it might be nice to have log-level confirmation of that too.

shiplet (approved):
approved with a non-blocking comment/suggestion!

great_expectations_cloud/agent/agent.py
alena-hutchinson: Good call. Added!

PR #624, opened 2025-01-10
Upgrade poetry in github workflows

anthonyburdi:
With poetry pinned and the caches invalidated, it looks like poetry 2.0.0 is working at least in the ci workflow. This PR upgrades poetry everywhere to 2.0.0, and after it merges if there is an issue with the pypi workflow we can downgrade there.


Shinnnyshinshin (approved):
👍🏼

PR #625, opened 2025-01-10
Publish to pypi python upgrade

anthonyburdi:
Bump the python version since poetry 2.0.0 does not support python 3.8 any longer https://python-poetry.org/blog/announcing-poetry-2.0.0#changed


Shinnnyshinshin (approved):
🚀

PR #633, opened 2025-01-16
Wait until the GX Agent's ThreadPoolExecutor has available thread before starting a job

alena-hutchinson:
This is the lowest-touch fix I could come up with to address the scheduler issues (see thread [here](https://greatexpectationslabs.slack.com/archives/C05R8RE43CL/p1736356433906939) for context)...

We wait until an instantaneous task completes before starting a job. When the instantaneous task completes, we know that the Agent's `ThreadPoolExecutor` has at least one available thread.

Obviously this needs some more testing, just looking for an initial temperature check. Ran locally and at a first pass, it seems like this is working as expected and allows us to plug in a higher value for more threads.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -7,7 +7,9 @@ import traceback
 import warnings
 from collections import defaultdict
 from concurrent.futures import Future
+from concurrent.futures import TimeoutError as CFTimeoutError
 from concurrent.futures.thread import ThreadPoolExecutor
+from datetime import timedelta
 from functools import partial
 from importlib.metadata import version as metadata_version
 from typing import TYPE_CHECKING, Any, Callable, Final, Literal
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -79,6 +81,7 @@
 # TODO Set in log dict
 LOGGER.setLevel(logging.DEBUG)
 HandlerMap = dict[str, OnMessageCallback]
+THREAD_TIMEOUT = timedelta(hours=2).total_seconds()
```
wookasz, line 84: Some of our jobs will undoubtedly take longer than this

alena-hutchinson: I just put this in here as a placeholder. I don't remember what we set the RabbitMQ heartbeat timeout to, but I think we'd want this value to be that minus some small buffer.

alena-hutchinson:
Don't want to do this because depending on what our prefectch count is, we could be blocking that many jobs from completing per worker.

PR #641, opened 2025-01-22
Update GX Core

wookasz:
Bump GX Core to 1.3.3

Also updating several other deps


poetry.lock
```diff
@@ -425,30 +425,43 @@ test-randomorder = ["pytest-randomly"]
 
 [[package]]
 name = "databricks-sql-connector"
-version = "3.7.1"
+version = "4.0.0b4"
```
NathanFarmer, line 428: Do you know why this is still here?

wookasz: Yes, it's a transitive dependency https://github.com/databricks/databricks-sqlalchemy/blob/main/pyproject.toml#L13

NathanFarmer (approved):

PR #660, opened 2025-02-04
Conditionally add expectations based on `selected_data_quality_issues`

alena-hutchinson:
Conditionally add expectations during `GenerateDataQualityCheckExpectationsAction.run()` based on `selected_data_quality_issues`. Also sets the `created_via` field on autogenerated expectations. 

Part of work for [Volume Change Detection](https://greatexpectations.atlassian.net/browse/GX-195). 

Tested this and it looks like things are working as expected. Screenshot below of VCD + SCD expectation. Also tested with just VCD checked. 
Screenshot 2025-02-04 at 4 30 42 PM


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -131,16 +148,34 @@ def _get_metrics(self, data_asset: DataAsset) -> tuple[MetricRun, UUID]:
...
 
+    def _add_volume_change_expectation(self, asset_id: UUID) -> UUID:
+        expectation = gx_expectations.ExpectTableRowCountToBeBetween(
+            windows=[
+                Window(
+                    constraint_fn="mean",
+                    parameter_name="min_value",
+                    range=1,
+                    offset=Offset(positive=0.0, negative=0.0),
+                )
```
NathanFarmer, line 159: My `strict` work introduced `strict` to `Window`. You will need to pass `strict=True` here to get the correct config/rendering. This is probably different than what was given in the ticket, but my work changed the config.

NathanFarmer (changes requested):
One tiny change!

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -131,16 +148,34 @@ def _get_metrics(self, data_asset: DataAsset) -> tuple[MetricRun, UUID]:
...
+            ],
+            strict_min=True,
```
NathanFarmer, line 161: You can probably omit `strict_min` here. Although this is more explicit so maybe it's preferred to do this.

alena-hutchinson: I'll leave it here to be explicit.

NathanFarmer (changes requested):
I think you want to delete `junit.xml`

NathanFarmer (approved):

PR #672, opened 2025-02-14
add user agent string

joshua-stauffer:
GX Core has recently added functionality which allows us to specify the the run environment for analytics. This PR updates the agent to use it.

### DIFF
--- a/great_expectations_cloud/agent/agent.py
+++ b/great_expectations_cloud/agent/agent.py
@@ -156,7 +156,9 @@ class GXAgent:
         with warnings.catch_warnings():
             # suppress warnings about GX version
             warnings.filterwarnings("ignore", message="You are using great_expectations version")
-            self._context: CloudDataContext = get_context(cloud_mode=True, user_agent_str="gx-agent")
+            self._context: CloudDataContext = get_context(
+                cloud_mode=True, user_agent_str="gx-agent"
+            )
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/agent.py
```diff
@@ -156,7 +156,9 @@ def __init__(self: Self):
         with warnings.catch_warnings():
             # suppress warnings about GX version
             warnings.filterwarnings("ignore", message="You are using great_expectations version")
-            self._context: CloudDataContext = get_context(cloud_mode=True)
+            self._context: CloudDataContext = get_context(
+                cloud_mode=True, user_agent_str="gx-agent"
```
tyler-hoffman, line 160: Classic bikeshedding on a one-line PR. The intended format (copied from the PR that added the value) is `user_agent_str: Optional string, should be of format <PRODUCT> / <VERSION> <COMMENT>` where comment is optional. Can we do `user_agent_str=f"gx-agent / {__version__}}"`?

dctalbot (approved):

tyler-hoffman (changes requested):
requesting an update to the form of the agent str

great_expectations_cloud/agent/agent.py
joshua-stauffer: sure

PR #682, opened 2025-03-03
[MAINTENANCE] Pass user_agent_str to get_context

tyler-hoffman:
We already send these in request headers from the agent, but for consistency with other repos, we're also initializing our analytics with it.


joshua-stauffer (approved):
👍

PR #702, opened 2025-03-19
Remove run window checkpoint action.

billdirks:
DO NOT MERGE UNTIL https://github.com/greatexpectationslabs/mercury/pull/4473 is MERGED

We removed the window checkpoint action since checkpoint.run in core now handles the window parameters.

I have verified locally that sending a checkpoint event with windowed events works with this PR.


great_expectations_cloud/agent/actions/run_window_checkpoint.py
wookasz, line 1: Should this file be getting deleted?

billdirks: Yes, I will delete. Thanks!

PR #713, opened 2025-03-24
add completeness expectations

shiplet:
Generates completeness expectations for each column on a given asset

Types of completeness expectations generated:
- if 100% of rows of a given column are null: `ExpectColumnValuesToBeNull(column=column, mostly=1)`
- if 100% of rows are not null: `ExpectColumnValuesToNotBeNull(column=column, mostly=1)`
- else: `ExpectColumnValuesToBeNull(windows=windows, column=column, mostly=parameter)`

https://github.com/user-attachments/assets/e3bcb1e7-e952-49a1-ba39-0fef6dd836c9

### DIFF
--- a/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
+++ b/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
@@ -34,7 +34,7 @@ from great_expectations_cloud.agent.models import (
 )
 from great_expectations_cloud.agent.utils import (
     TriangularInterpolationOptions,
-    faker_string_alpha,
+    param_safe_unique_id,
     triangular_interpolation,
 )
 
@@ -148,7 +148,7 @@ class GenerateDataQualityCheckExpectationsAction(
 
         except Exception as e:
             # TODO - see if this can be made more specific
-            raise RuntimeError(f"Failed to retrieve asset: {e}") from e  # noqa: TRY003 # want to keep this informative for now
+            raise RuntimeError(f"Failed to retrieve asset: {e}") from e  # noqa: TRY003 keep this informative for now
 
         return data_asset
 
@@ -172,7 +172,7 @@ class GenerateDataQualityCheckExpectationsAction(
         return metric_run, metric_run_id
 
     def _add_volume_change_expectation(self, asset_id: UUID) -> UUID:
-        unique_id = faker_string_alpha(16)
+        unique_id = param_safe_unique_id(16)
         expectation = gx_expectations.ExpectTableRowCountToBeBetween(
             windows=[
                 Window(
@@ -215,7 +215,7 @@ class GenerateDataQualityCheckExpectationsAction(
 
     def _add_completeness_change_expectations(
         self, metric_run: MetricRun, asset_id: UUID
-    ) -> [UUID]:
+    ) -> list[UUID]:
         table_row_count = next(
             metric
             for metric in metric_run.metrics
@@ -248,7 +248,7 @@ class GenerateDataQualityCheckExpectationsAction(
                     column=column_name, mostly=1
                 )
             else:
-                unique_id = faker_string_alpha(16)
+                unique_id = param_safe_unique_id(16)
                 options = TriangularInterpolationOptions(
                     input_range=(0.0, float(row_count)), output_range=(0, 1), round_precision=3
                 )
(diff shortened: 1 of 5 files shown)


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -195,6 +213,69 @@ def _add_schema_change_expectation(self, metric_run: MetricRun, asset_id: UUID)
...
+                    column=column_name, mostly=1
+                )
+            elif null_count == row_count:
+                expectation = gx_expectations.ExpectColumnValuesToBeNull(
+                    column=column_name, mostly=1
+                )
+            else:
+                unique_id = param_safe_unique_id(16)
+                options = TriangularInterpolationOptions(
+                    input_range=(0.0, float(row_count)), output_range=(0, 1), round_precision=3
```
ashmortar, line 253: I think the output range we want here is 0-10?  It depends somewhat on what mostly takes.  Ideally we want to output something that will correspond to a mostly value between 0 and 10%

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -195,6 +213,69 @@ def _add_schema_change_expectation(self, metric_run: MetricRun, asset_id: UUID)
...
+                )
+                interpolated_offset = triangular_interpolation(null_count, options)
+                expectation = gx_expectations.ExpectColumnValuesToBeNull(
+                    windows=[
+                        Window(
+                            constraint_fn="mean",
+                            parameter_name=f"{unique_id}_null_value_min",
+                            range=1,
```
ashmortar, line 261: And we will want over the last 5 runs, not sure if this is  where that is set

ashmortar (changes requested):
Couple of questions/changes.  It could be my ignorance of how the windowed expectations are created.

I think ideally we would also check in tests that the expectations we create have the expected values for the windowed params.

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
shiplet: ah, okay good to know. I thought it was normalized from 0 to 1. Will update.

shiplet: @ashmortar Hmm. Updating from 0 to 10 moves the resulting mostly percentages up by an order of magnitude as well. So originally the percent deviation on some would be 3%, and going from 0 to 10 shifts them to 30%.

ashmortar: Ah, in that case you were correct 😅

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -195,6 +213,69 @@ def _add_schema_change_expectation(self, metric_run: MetricRun, asset_id: UUID)
...
+        for column in column_null_values_metric:
+            column_name = column.column
+            null_count = column.value
+            row_count = table_row_count.value
+            if null_count == 0:
+                expectation = gx_expectations.ExpectColumnValuesToNotBeNull(
+                    column=column_name, mostly=1
+                )
+            elif null_count == row_count:
+                expectation = gx_expectations.ExpectColumnValuesToBeNull(
```
ashmortar, line 247: Sorry, one more change requested here, I think we want to use the same expectation `ExpectColumnValuesToNotBeNull` but with mostly being 0

ashmortar (changes requested):

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
shiplet: Ah, so the reason I differentiated between `ToBeNull` and `ToNotBeNull` was down to the renderer. If we make them `ToNotBeNull` with `mostly=0`, they render like 
image

shiplet: sounds good - and if we see it in practice and decide to change it, it's not difficult to change too

ashmortar (approved):

PR #734, opened 2025-04-04
add draft configs to event

shiplet:


josectobar (approved):

great_expectations_cloud/agent/models.py
```diff
@@ -120,6 +120,7 @@ class RunRdAgentEvent(EventBase):
     batch_definition_name: str
     batch_parameters: Optional[dict[str, Any]] = None
     use_core_metrics: bool = False
+    create_draft_configs: bool = False
```
josectobar, line 123:
```suggestion
    create_expectation_draft_configs: bool = False
```

josectobar: I had the same question in the other PR. Do we need to say `create`?

PR #738, opened 2025-04-09
install gx-redshift

dctalbot:


pyproject.toml
```diff
@@ -37,6 +37,7 @@ snowflake-sqlalchemy = { version = ">=1.6", optional = true }
 snowflake-connector-python = { version = ">=3.3.1", optional = true }
 psycopg2-binary = { version = "*", optional = true }
 databricks-sqlalchemy = { version = ">=1.0.0", optional = true }
+sqlalchemy-redshift = { git = "https://github.com/great-expectations/sqlalchemy-redshift.git" }
```
dctalbot, line 40: @billdirks I think it would be preferable to install great-expectations with the gx-redshift extra, but I'll try this until that extra is available

PR #748, opened 2025-04-15
refactor: requirements

dctalbot:
- ⚠️  This change breaks certain developer workflows:
  - all poetry extras were removed and their members are now installed by default
  - the `sql` poetry group was removed (extraneous)

Also:
- install SQL dialects via `great-expectations` extras instead of transitive dependencies
- remove `contributing.md` because it did not include any original material
- DRY up the CI python setup, use pinned poetry version


.github/workflows/ci.yaml
```diff
@@ -14,69 +14,16 @@ jobs:
...
-      - name: Install Poetry
-        if: steps.cached-poetry.outputs.cache-hit != 'true'
-        uses: snok/install-poetry@v1
-        with:
-          version: 2.0.0
-
-      - name: Install dependencies
-        run: poetry install --sync
-
+      - uses: actions/checkout@v4
```
wookasz, line 25: Is `Checkout` the default name for this one?

wookasz (approved):

PR #758, opened 2025-04-24
Upgrade GX Core

NathanFarmer:
- Upgrade GX Core to version `1.4.2`
- Fix mypy errors due to newly exported typing information


tests/agent/actions/test_generate_data_quality_check_expectations_action.py
```diff
@@ -291,9 +295,6 @@ def test_missing_table_columns_metric_raises_runtime_error(
     mock_data_asset.id = uuid.uuid4()
     mock_data_asset.test_connection.return_value = None
 
-    # Mock the context to return our mock data asset
-    mock_context.data_sources.get.return_value.get_asset.return_value = mock_data_asset
-
```
NathanFarmer, line 296: Chaining `return_value` doesn't seem to work like this, and removing this line made no difference regardless.

tests/agent/actions/test_run_metrics_list_action.py
```diff
@@ -94,7 +94,7 @@ def test_run_metrics_list_computes_metric_run_missing_batch_inspector(
     )
 
     # Asserting against mocked class proved challenging
-    action._batch_inspector.compute_metric_list_run.assert_called_once()
+    action._batch_inspector.compute_metric_list_run.assert_called_once()  # type: ignore[attr-defined]  # this is not a mock
```
NathanFarmer, line 97: I considered removing this line, but that would mean removing the entire test since this is the only assertion. It seems to work but the type checker complains that there is no `assert_called_once()` 🤔

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -150,12 +154,12 @@ def _retrieve_asset_from_asset_name(
...
 
-        return data_asset
+        return data_asset  # type: ignore[no-any-return]  # unable to narrow types strictly based on names
 
-    def _get_metrics(self, data_asset: DataAsset) -> tuple[MetricRun, UUID]:
+    def _get_metrics(self, data_asset: DataAsset[Any, Any]) -> tuple[MetricRun, UUID]:
         batch_request = data_asset.build_batch_request()
         metric_run = self._batch_inspector.compute_metric_list_run(
-            data_asset_id=data_asset.id,
+            data_asset_id=data_asset.id or uuid4(),  # id can be None
```
billdirks, line 162: Is it every None in practice? That is, we didn't do this before; did we use to crash? I'm wondering if it's better if we check and raise if it's None.

great_expectations_cloud/agent/actions/run_checkpoint.py
```diff
@@ -64,7 +64,7 @@ def run_checkpoint(
     created_resources = []
     for key in validation_results.keys():
         created_resource = CreatedResource(
-            resource_id=validation_results[key].id,
+            resource_id=validation_results[key].id or "",  # id can be None
```
billdirks, line 67: Same comment as before. Is this ever None in practice?

great_expectations_cloud/agent/agent.py
```diff
@@ -97,8 +97,7 @@ class GXAgentConfig(AgentBaseExtraForbid):
 
     queue: str
     connection_string: AmqpDsn
-    # pydantic will coerce this string to AnyUrl type
-    gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL
+    gx_cloud_base_url: AnyUrl = AnyUrl(url=CLOUD_DEFAULT_BASE_URL, scheme="https")
```
billdirks, line 100: Is hardcoding the scheme going to make this hard to test locally? If pydantic is going to coerce this, maybe we should let it and put in a type ignore?

great_expectations_cloud/agent/config.py
```diff
@@ -7,8 +7,7 @@
 
 
 class GxAgentEnvVars(BaseSettings):
-    # pydantic will coerce this string to AnyUrl type
-    gx_cloud_base_url: AnyUrl = CLOUD_DEFAULT_BASE_URL
+    gx_cloud_base_url: AnyUrl = AnyUrl(url=CLOUD_DEFAULT_BASE_URL, scheme="https")
```
billdirks, line 10: Same as above.

tests/agent/actions/test_run_metrics_list_action.py
billdirks: If it's not a mock, do we define this method somewhere?

tests/integration/actions/test_metric_list_action.py
```diff
@@ -92,16 +93,17 @@ def execute(self, query, variables=None, headers=None, status_code=None):
...
     datasource = context.data_sources.get(name=datasource_name)
+    assert isinstance(datasource, PostgresDatasource)
     yield datasource
 
 
 @pytest.fixture
 def local_mercury_db_organizations_table_asset(
-    local_mercury_db_datasource: PostgresDatasource,
-) -> TableAsset:
+    local_mercury_db_datasource: Any,
```
billdirks, line 105: Instead of `Any`, can you match the return type of the fixture above? Maybe you have to change it a  little since it's a Generator.

alena-hutchinson (approved):
Wow, this is a ton of cleanup. Thanks for doing this! LGTM

billdirks (commented):
Thank you! I left a few small comments. The one I'm concerned about is the AnyUrl and forcing the scheme to be https since I'm wondering if that will impact local testing.

shiplet (approved):
This looks good - did you get a chance to do any kind of smoke-testing with it locally?

great_expectations_cloud/agent/actions/run_checkpoint.py
NathanFarmer: Probably not. I'm not sure.

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
NathanFarmer: I'm not sure. Maybe in this case it's better to raise.

great_expectations_cloud/agent/agent.py
NathanFarmer: Why would it make this hard to test? I could move the `AnyUrl` object into the constant. We could let it coerce, but I'm not sure I understand what the reason for doing that would be.

tests/agent/actions/test_run_metrics_list_action.py
NathanFarmer: No, that's what the type error is about. I could try to figure out how to mock it, but I'm leaning towards removing this test since I don't understand what it's testing.

great_expectations_cloud/agent/agent.py
NathanFarmer: The string constant `CLOUD_DEFAULT_BASE_URL` already starts with `https`. `scheme` is just a required field for `AnyUrl`.

PR #767, opened 2025-04-30
Temporary fix to handle CCD error when column is empty

alena-hutchinson:
This is a band aid solution so that we don't fail with [this](https://gx-cloud.sentry.io/issues/6509381994/?referrer=jira_integration) Sentry when a column is empty and a user wants to generate CCD expectations. We should figure out why the metric is being returned as `None` instead of `0`, but for now this will restore functionality to the UI.

### DIFF
--- a/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
+++ b/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
@@ -271,6 +271,10 @@ class GenerateDataQualityCheckExpectationsAction(
                 options = TriangularInterpolationOptions(
                     input_range=(0.0, float(row_count)), output_range=(0, 0.1), round_precision=5
                 )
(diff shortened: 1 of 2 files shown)


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -271,6 +271,10 @@ def _add_completeness_change_expectations(
                 options = TriangularInterpolationOptions(
                     input_range=(0.0, float(row_count)), output_range=(0, 0.1), round_precision=5
                 )
+                if null_count is None:
```
billdirks, line 274: I think we should move this up into another `elif` and not make any expectations in this case. `None` indicates not rows in the table so we don't have any expectations about nullity.

billdirks (approved):
Thanks!

PR #787, opened 2025-05-23
Fetch views when listing assets in a database

wookasz:
We often recommend using views in various use cases. This allows these views to be added via the UI

### DIFF
--- a/great_expectations_cloud/agent/actions/utils.py
+++ b/great_expectations_cloud/agent/actions/utils.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
     from sqlalchemy.engine import Inspector
 
 
-def get_table_names(datasource: SQLDatasource) -> list[str]:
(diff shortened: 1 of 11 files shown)


great_expectations_cloud/agent/actions/utils.py
```diff
@@ -9,14 +9,18 @@
...
         # Also converting to list to ensure JSON serializable
-        return list(inspector.get_table_names(schema=datasource.schema_))
+        tables = list(inspector.get_table_names(schema=datasource.schema_))
+        views = list(inspector.get_view_names(schema=datasource.schema_))
+        return tables + views
 
-    return list(inspector.get_table_names())
+    tables = list(inspector.get_table_names())
+    views = list(inspector.get_view_names())
+    return tables + views
```
wookasz, line 26: this is the only functionality change in this PR

PR #802, opened 2025-06-03
Only add schema and volume Anomaly Detection coverage to an Asset if it isn't already covered

alena-hutchinson:
Leverage the new `expectation_v1` [endpoint](https://github.com/greatexpectationslabs/mercury/pull/4638/files) so that we only add schema and volume change Anomaly Detection Expectations if the Asset does not already have coverage. 

**DO NOT MERGE until `mercury` PR linked above has been deployed`**


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -177,6 +188,30 @@ def _get_metrics(self, data_asset: DataAsset[Any, Any]) -> tuple[MetricRun, UUID
 
         return metric_run, metric_run_id
 
+    def _get_current_anomaly_detection_coverage(self, asset_id: UUID | None) -> dict[str, Any]:
```
billdirks, line 191: Can you replace Any with something more specific?

tests/agent/actions/test_generate_data_quality_check_expectations_action.py
```diff
@@ -76,12 +76,20 @@ def mock_data_asset(self, event: GenerateDataQualityCheckExpectationsEvent, asse
     def mock_metrics(self, data_asset: DataAsset[Any, Any]):
         return MetricRun(metrics=mock_metrics_list), uuid.uuid4()
 
+    def mock_no_anomaly_detection_coverage(self, data_asset: DataAsset[Any, Any]):
```
billdirks, line 79: Can you add some tests where we return some anomaly detection expectations and verify we don't create new ones?

billdirks (commented):
Thanks for doing this! I've added some comments/questions inline.

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -94,39 +96,50 @@ def run(self, event: GenerateDataQualityCheckExpectationsEvent, id: str) -> Acti
...
+                    if DataQualityIssues.COMPLETENESS in selected_dqi:
+                        completeness_change_expectation_ids = (
+                            self._add_completeness_change_expectations(
+                                metric_run=metric_run, asset_id=data_asset.id
+                            )
                         )
+                        for exp_id in completeness_change_expectation_ids:
+                            created_resources.append(
+                                CreatedResource(resource_id=str(exp_id), type="Expectation")
+                            )
```
alena-hutchinson, line 143: This diff looks worse than it actually is- the large diff is because the existing code block was pulled into an `if` statement to check if any DQI's were selected, so that we don't run an expensive query to fetch the Expectations if we don't need to

great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -94,39 +96,50 @@ def run(self, event: GenerateDataQualityCheckExpectationsEvent, id: str) -> Acti
...
                     )
-                    created_resources.append(
-                        CreatedResource(
-                            resource_id=str(schema_change_expectation_id), type="Expectation"
+
+                    if (
+                        DataQualityIssues.VOLUME in selected_dqi
+                        and DataQualityIssues.VOLUME not in pre_existing_anomaly_detection_coverage
+                    ):
+                        constraint_fn = "forecast" if event.use_forecast else "mean"
```
billdirks, line 109: Can we define an enum for the constraint function types with `forecast` and `mean` and then use that throughout this file.

billdirks (changes requested):
Thanks for doing this! I left a few suggestions inline. It would be good to have an enum for the constraint functions and to do some validation of the GET response.

PR #818, opened 2025-06-13
Remove reliance on Data Source names when running Checkpoints

wookasz:
## Why
Schedules are not updated when a DS name changes which breaks schedules. The Checkpoint, however, always has the correct DS name injected into it, so rather than relying on the DS and asset names in the payload, we can use the ones stored in the Checkpoint.

### DIFF
--- a/great_expectations_cloud/agent/actions/run_checkpoint.py
+++ b/great_expectations_cloud/agent/actions/run_checkpoint.py
@@ -59,16 +59,17 @@ def run_checkpoint(
     for vd in checkpoint.validation_definitions:
         ds = vd.data_source
         ds_name = ds.name
+        # create assets by name dict
         if ds_name not in data_sources_assets_by_data_source_name:
             data_sources_assets_by_data_source_name[ds_name] = DataSourceAssets(
                 data_source=ds, assets_by_name={}
             )
         data_sources_assets_by_data_source_name[ds_name].assets_by_name[vd.asset.name] = vd.asset
 
-    for _, data_sources_assets in data_sources_assets_by_data_source_name.items():
+    for data_sources_assets in data_sources_assets_by_data_source_name.values():
         data_source = data_sources_assets.data_source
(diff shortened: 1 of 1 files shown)


tyler-hoffman (approved):
Left minor suggestions, but LGTM

great_expectations_cloud/agent/actions/run_checkpoint.py
```diff
@@ -30,32 +32,45 @@ class MissingCheckpointNameError(ValueError):
...
+    for vd in checkpoint.validation_definitions:
+        ds = vd.data_source
+        ds_name = ds.name
+        if ds_name not in data_sources_assets_by_data_source_name:
+            data_sources_assets_by_data_source_name[ds_name] = DataSourceAssets(
+                data_source=ds, assets_by_name={}
+            )
+        data_sources_assets_by_data_source_name[ds_name].assets_by_name[vd.asset.name] = vd.asset
+
+    for _, data_sources_assets in data_sources_assets_by_data_source_name.items():
```
NathanFarmer, line 68:
```suggestion
    for data_sources_assets in data_sources_assets_by_data_source_name.values():
```

great_expectations_cloud/agent/actions/run_checkpoint.py
```diff
@@ -30,32 +32,45 @@ class MissingCheckpointNameError(ValueError):
...
+        data_source = data_sources_assets.data_source
+        data_source.test_connection(test_assets=False)  # raises `TestConnectionError` on failure
+        for _, data_asset in data_sources_assets.assets_by_name.items():
```
NathanFarmer, line 71:
```suggestion
        for data_asset in data_sources_assets.assets_by_name.values():
```

NathanFarmer (changes requested):
You do know why we are doing `data_source.test_connection(test_assets=False)` and then `data_asset.test_connection()` immediately after?

PR #822, opened 2025-06-13
Only add completeness expectations to columns without coverage

alena-hutchinson:
Only add completeness change Expectations to columns that are missing coverage. 

This behavior is behind the `expect_non_null_proportion_enabled` feature flag.

### DIFF
--- a/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
+++ b/great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
@@ -513,10 +513,10 @@ class GenerateDataQualityCheckExpectationsAction(
     ) -> list[ColumnMetric[int]]:
         try:
             columns_with_completeness_coverage = {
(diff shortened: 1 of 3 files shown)


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -112,10 +112,9 @@ def run(self, event: GenerateDataQualityCheckExpectationsEvent, id: str) -> Acti
                         self._get_current_anomaly_detection_coverage(data_asset.id)
                     )
 
-                    if self._should_add_anomaly_detection_coverage_for_issue(
-                        DataQualityIssues.VOLUME,
-                        selected_dqis,
-                        pre_existing_anomaly_detection_coverage,
+                    if self._should_add_volume_change_coverage(
```
alena-hutchinson, line 115: Changed this to be more clear/specific, since the original function isn't one-size-fits-all anyway.

josectobar (approved):
This looks good to me.

PR #826, opened 2025-06-17
Remove `expectNonNullProportionEnabled` feature flag and `expect_non_null_proportion_enabled` parameter

NathanFarmer:


tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -297,20 +297,20 @@ def mock_no_anomaly_detection_coverage(self, data_asset_id: uuid.UUID | None):
...
+            == "expect_column_proportion_of_non_null_values_to_be_between"
+        )
         # Verify each has required properties
         assert exp_config.column is not None
         if exp_config.windows is not None:
-            assert exp_config.windows[0].constraint_fn == "mean"
-            assert exp_config.windows[0].range == 5
-            assert exp_config.windows[0].offset.positive == exp_config.windows[0].offset.negative
-            assert exp_config.windows[0].parameter_name == exp_config.mostly["$PARAMETER"]
+            for window in exp_config.windows:
```
tyler-hoffman, line 307: We should not have that `if`. It's unclear if we even make it to this assertion. We should also assert that windows not empty for similar reasons.

tyler-hoffman (approved):
Left two issues with one of the tests. One issue predates this PR, and one was introduced here. I'd recommend updating here, but not strictly blocking.

tests/integration/actions/test_generate_data_quality_check_expectations.py
NathanFarmer: The issue is that this is an integration test that uses real data. Depending on the state of the columns in the table, some Expectations might not have windows, and others might. The Expectations that have windows would have this configuration.

PR #830, opened 2025-06-18
Make `Event` system extensible

NathanFarmer:
Build dynamic `Union` type for `Event` at runtime


tests/agent/test_models.py
```diff
@@ -0,0 +1,216 @@
+from __future__ import annotations
+
+import json
+from typing import Any, Literal
+from uuid import UUID
+
+import pytest
+from pydantic.v1 import ValidationError, parse_raw_as
```
NathanFarmer, line 8: I guess this whole codebase is still on `pydantic.v1`

ashmortar (approved):

PR #833, opened 2025-06-23
Fredwang1012

fredwang1012:


great_expectations_cloud/agent/actions/run_checkpoint.py
```diff
@@ -50,28 +56,47 @@ def run_checkpoint(
...
+    if hasattr(event, 'datasource_names_to_asset_names') and event.datasource_names_to_asset_names:
+        # Use your original approach for backward compatibility
+        for datasource_name, data_asset_names in event.datasource_names_to_asset_names.items():
+            datasource = context.data_sources.get(name=datasource_name)
+            datasource.test_connection(test_assets=False)  # raises `TestConnectionError` on failure
+            for data_asset_name in data_asset_names:
+                asset = datasource.get_asset(data_asset_name)
+                verify_data_asset(LOGGER.info, datasource, asset)
+                asset.test_connection()  # raises `TestConnectionError` on failure
+    else:
```
wookasz, line 69: This will always be true. We left `datasource_names_to_asset_names` in the message payload for backward compatibility with older agent versions.

The approach of using validation definitions should always work. Was there something not working for you?

great_expectations_cloud/agent/actions/utils.py
```diff
@@ -20,7 +23,22 @@ def get_asset_names(datasource: SQLDatasource) -> list[str]:
...
     return tables + views
+
+def verify_data_asset(log, datasource: SQLDatasource, asset):
+    try:
+        engine = datasource.get_engine()
+        if engine.dialect.name == GXSqlDialect.MSSQL and asset.type == "table" and asset.schema_name == None and ("." in asset.name):
+            datasource.delete_asset(asset.name)
+            datasource.add_table_asset(name = asset.name, schema_name = asset.name.split('.')[0], table_name = asset.name.split('.')[1])
+    except Exception as e:
+        log(f"Failed to verify asset {asset.name}: {e}")
```
wookasz, line 44: Deleting and re-creating the asset should not happen as part of the actions for a couple reasons:
- the newly created asset is actually an entirely new entity, with a new UUID. Only the names are the same
- recreating the asset means that new results are no longer associated with the validation history of the old asset


Is the intent of this method to ensure that the asset is created if it was for some reason deleted?

great_expectations_cloud/agent/actions/utils.py
```diff
@@ -20,7 +23,22 @@ def get_asset_names(datasource: SQLDatasource) -> list[str]:
         tables = list(inspector.get_table_names(schema=datasource.schema_))
         views = list(inspector.get_view_names(schema=datasource.schema_))
         return tables + views
+    
+    engine = datasource.get_engine()
+    if engine.dialect.name == GXSqlDialect.MSSQL:
+        tables = list(chain.from_iterable(list(map(lambda o: f"{schema}.{o}", inspector.get_table_names(schema=schema))) for schema in inspector.get_schema_names()))
+        views = list(chain.from_iterable(list(map(lambda o: f"{schema}.{o}", inspector.get_view_names(schema=schema))) for schema in inspector.get_schema_names()))
```
wookasz, line 30: Can you explain the intent of these lines?

The Data Source configuration should have schema configured on it and with our other connectors, that configuration is used on the GX Core side when performing operations against the table. Is that not working for MS SQL? If that's the case, it seems we should be implementing a fix in GX Core rather than in the agent

wookasz (changes requested):

great_expectations_cloud/agent/actions/utils.py
fredwang1012: These lines intend to get all the tables and views in a schema in a database. Our SQL Server is structured as follows: [database].[schema].[table]. So, the idea here is to get all of the tables in a datasource by looping through all of the schemas in that datasource. I think you are right about the data source configurations, though, meaning we should have [database].[schema] as a data source instead of a whole database. I will set up a data source like that, and if it works, we can remove this bit of code. There is a way to add schema into the connection string for SQL Server, right?

great_expectations_cloud/agent/actions/utils.py
fredwang1012: This serves to properly set up a newly created data asset for SQL Server. When we currently add an asset on GX Cloud, it is in the form of [schema].[table], which is treated like just a table. This was just a hacky way of intercepting the data asset addition and verifying the asset as [schema].[table]. But, similar to your comment above, I think if we just set up the data source as a schema, we probably won't need this code.

great_expectations_cloud/agent/actions/run_checkpoint.py
fredwang1012: I will test it out again and see if we need the conditional or not.

great_expectations_cloud/agent/actions/utils.py
wookasz: Looking at Microsoft's doc, looks like Schema might not be supported for SQL Server :( At least I do not see any mention of it in the documents. If that's the case, I think there are 2 options:
1. taking your approach of returning the concatenated names
2. adding logic into GX Core to do this concatenation at runtime

I'd like to avoid (1) to keep data source-specific logic in GX Core, as much as possible, and also to ensure GX Core maintains consistent behavior across data sources.

Let me look into the options here and get back to you

fredwang1012: Hey @wookasz, I will try out option two and see if I can figure something out.

wookasz: @fredwang1012 I think there are already some patterns in the codebase that could be borrowed here.

- add_table_asset have an optional `schema_name` parameter. Refer to our `SQLDatasource` [documentation](https://docs.greatexpectations.io/docs/reference/api/datasource/fluent/SQLDatasource_class) for an example.
- The `SnowfalkeDatasource` has a `schema_` method [here](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/snowflake_datasource.py#L356-L359).
- `SnowflakeDatasource` also overrides `add_table_asset` to inject the schema name from the data source if one is not specified. See [here](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/snowflake_datasource.py#L532).
- Throughout the rest of the codebase, when an Asset has a `schema_name` set, we resolve its `qualifed_name` as `<schema_name.table_name>`. See [here](https://github.com/great-expectations/great_expectations/blob/develop/great_expectations/datasource/fluent/sql_datasource.py#L1008-L1010).

A good option here would be to create a `SQLServerDatasource` subclass and adding overrides similar to those for `SnowflakeDatasource`. This would maintain consistency between implementations.

I believe this should also be compatible with how our backend and UI expect Assets to be configured, but once the changes are in core I can test and make changes there if necessary.

PR #839, opened 2025-06-30
Do not set autogenerated or created_via fields in action

alena-hutchinson:
We no longer want to set the `autogenerated` or `created_via` field during the `GenerateDataQualityCheckExpectationsAction` since we can now trigger this action through the "New Expectation" workflow. We do not use these fields in the FE, and have discussed the impact to analytics [here](https://greatexpectationslabs.slack.com/archives/C08K6HWULR0/p1751313279409839)


josectobar (approved):
This looks good to me. Do we need to wait before we merge this since we can't put it behind a FF?

alena-hutchinson:
> This looks good to me. Do we need to wait before we merge this since we can't put it behind a FF?

I think we are okay since we already merged the FE PR which was looking at these fields.

PR #861, opened 2025-07-30
support forecasting completeness checks

dctalbot:
- upon asset creation, generate a completeness check that uses a forecasting model to generative the values for the bounds at runtime


NathanFarmer (approved):

tests/agent/actions/test_generate_data_quality_check_expectations_action.py
```diff
@@ -759,6 +759,91 @@ def mock_create_expectation(expectation, asset_id, created_via):
...
+            schema_name="test_schema",
+        )
+
+    def mock_get_metrics(data_asset):
+        return MetricRun(metrics=metrics), uuid.uuid4()
+
+    def mock_get_coverage(data_asset_id):
+        return {}
+
+    mocker.patch.object(action, "_retrieve_asset_from_asset_name", side_effect=mock_retrieve_asset)
```
tyler-hoffman, line 797: not blocking, but I'd recommend against mocking private methods since they should be treated as an implementation detail. I don't think it matters enough to spend too many cycles on this though.

tests/agent/actions/test_generate_data_quality_check_expectations_action.py
```diff
@@ -759,6 +759,91 @@ def mock_create_expectation(expectation, asset_id, created_via):
...
+    mocker.patch.object(action, "_get_metrics", side_effect=mock_get_metrics)
+    mocker.patch.object(
+        action, "_get_current_anomaly_detection_coverage", side_effect=mock_get_coverage
+    )
+
+    # Mock the _create_expectation_for_asset method to capture created expectations
+    created_expectations = []
+
+    def mock_create_expectation(expectation, asset_id, created_via):
+        created_expectations.append(expectation)
```
tyler-hoffman, line 807: total nit: probably cleaner to just assert what a spy was called with rather than implement a mechanism to capture these

tyler-hoffman (approved):
Left nits, but feel free to ignore. LGTM

tests/agent/actions/test_generate_data_quality_check_expectations_action.py
dctalbot: Agreed, this was a pattern that I copied so I'll keep it as-is for now. I appreciate the note.

PR #870, opened 2025-08-07
Add warning level severity to automatic Expectations

alena-hutchinson:
Automatically generated Expectations should be created with a default failure `severity` of `warning`. The `severity` field lives in the Expectation's `kwargs` dict. 

This PR also has some test cleanup so that we can run the `test_generate_data_quality_check_expectaitons.py` integration test locally. Specifically, we were overwriting the `GX_CLOUD_BASE_URL` environment variable in `conftest.py`, and running into DB constraint violations due to the way data was being seeded/cleaned up in these tests.


tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -32,64 +33,98 @@ def user_api_token_headers_org_admin_sc_org():
     }
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
```
alena-hutchinson, line 36: This makes it so we only seed/teardown data one time, not before/after each test. Previously we would hit race conditions between cleanup/teardown during test execution.

tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -32,64 +33,98 @@ def user_api_token_headers_org_admin_sc_org():
...
 def seed_and_cleanup_test_data(context: CloudDataContext):
     # Seed data
     data_source_name = "local_mercury_db"
     data_source = context.data_sources.get(data_source_name)
     assert isinstance(data_source, PostgresDatasource)
 
-    table_data_asset = data_source.add_table_asset(
-        table_name="checkpoints", name="local-mercury-db-checkpoints-table"
-    )
+    table_data_asset_name = "local-mercury-db-checkpoints-table"
```
alena-hutchinson, line 43: The rest of this diff looks worse than it actually is. It adds checks so that we don't try to create/update a resource that already exists.

shiplet (approved):
Thanks for fighting through this! As a future improvement we may want to consider implementing session-scoped & batched setups and teardowns like we have in mercury.

PR #887, opened 2025-08-23
Automatic Expectations are created with warning-level severity 

alena-hutchinson:
Automatic Expectations are created with warning-level severity. The `severity` field now lives in the top-level of the `ExpectationConfig`, ie. `Expectation` > `config` instead of within the `kwargs`. 

This PR also does some integration test cleanup so that the relevant test file can be run locally. There were issues with seeding/cleaning up test data and env variables.


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -524,6 +524,8 @@ def _create_expectation_for_asset(
         expectation_payload["autogenerated"] = True
         if created_via is not None:
             expectation_payload["created_via"] = created_via
+        # Automatic expectations are created with warning-level severity
+        expectation_payload["severity"] = "warning"
```
wookasz, line 528: Since `Expectations` have a `severity` parameter, rather than injecting a severity into the payload, it should be passed when the expectation is being created.

That change would have to wait until GX Core goes out.

tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -32,64 +33,95 @@ def user_api_token_headers_org_admin_sc_org():
     }
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
```
wookasz, line 36: Why is this being made module level?

tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -32,64 +33,95 @@ def user_api_token_headers_org_admin_sc_org():
...
 def seed_and_cleanup_test_data(context: CloudDataContext):
     # Seed data
     data_source_name = "local_mercury_db"
     data_source = context.data_sources.get(data_source_name)
     assert isinstance(data_source, PostgresDatasource)
 
-    table_data_asset = data_source.add_table_asset(
-        table_name="checkpoints", name="local-mercury-db-checkpoints-table"
-    )
+    table_data_asset_name = "local-mercury-db-checkpoints-table"
```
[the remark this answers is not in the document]
wookasz: It seems like that's a symptom of something with how we run tests. Are we not doing proper cleanup?

If so, try finalizers.

If it's tests interfering with each other, adding randomness to the entity names may be another option.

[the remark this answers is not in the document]
wookasz: 🤯 this fixture looks like it needs a test

tests/integration/actions/test_generate_data_quality_check_expectations.py
```diff
@@ -32,64 +33,95 @@ def user_api_token_headers_org_admin_sc_org():
...
+    with engine.connect() as conn:
+        # Check if there's already a gx_managed=true validation for this asset
+        existing_gx_managed_validations = conn.execute(
+            sa.text(
+                f"SELECT COUNT(*) FROM validations WHERE asset_ref_id='{table_data_asset.id}' AND gx_managed=true"
+            )
+        ).scalar()
+
+        # Only set gx_managed=true if there isn't already one for this asset
+        if not existing_gx_managed_validations:
```
wookasz, line 88: This indicates that the environment is not deterministic. If there are conditionals in the test fixtures, I can't trust what the tests are doing

wookasz (changes requested):

tests/integration/actions/test_generate_data_quality_check_expectations.py
alena-hutchinson: Reverted these changes. I didn't need to end up making any changes to the integration tests, they are all in the unit test and I was able to run the integration test locally.

wookasz (approved):

PR #926, opened 2025-09-17
Add DomainContext to event_handler and agent_action initializer

josectobar:


tests/agent/actions/test_draft_datasource_config.py
```diff
@@ -79,7 +80,7 @@ def test_test_draft_datasource_config_success_non_sql_ds(
     event = DraftDatasourceConfigEvent(
         config_id=config_id,
         organization_id=uuid.uuid4(),
-        workspace_id=uuid.uuid4(),
+        workspace_id=workspace_id,
```
billdirks, line 83: It seems strange that workspace id is a predetermined value but organization id is a random uuid. Should we have `organization_id=org_id`? Maybe it's never used in this in this test but it seems like it should logically be set to this?
There are similar instances below.

billdirks (approved):
Thanks for doing this! It is much better. I have one question inline.

PR #1153, opened 2026-04-10
GX-3007: selective metric list in _get_metrics based on selected DQIs

ashmortar:
The _get_metrics method previously always fetched the same fixed set of profiling metrics regardless of which Data Quality Issues the user selected. This caused unnecessary computation — for example, fetching column non-null counts even when Completeness AD was not requested.

This PR adds a selected_dqis argument to _get_metrics and builds the metric list selectively: schema metrics (TABLE_COLUMNS, TABLE_COLUMN_TYPES) only when SCHEMA is selected, completeness metrics (TABLE_ROW_COUNT, COLUMN_NON_NULL_COUNT) only when COMPLETENESS is selected, and COLUMN_UNIQUE_PROPORTION only when UNIQUENESS is selected. Volume expectations use historical windows rather than current metric values, so VOLUME requires no profiling metrics.

Unblocks GX-3008.


great_expectations_cloud/agent/actions/generate_data_quality_check_expectations_action.py
```diff
@@ -238,19 +238,66 @@ def _retrieve_asset_from_asset_name(
...
+    ) -> tuple[MetricRun, UUID]:
         batch_request = data_asset.build_batch_request()
         if data_asset.id is None:
             raise RuntimeError("DataAsset.id is None")  # noqa: TRY003
-        metric_run = self._batch_inspector.compute_metric_list_run(
-            data_asset_id=data_asset.id,
-            batch_request=batch_request,
-            metric_list=[
+
+        if not selected_dqis:
```
alena-hutchinson, line 250: is this code we will eventually remove?

ashmortar: I wasn't totally sure if there were callers to `_get_metrics` that wouldn't be supplying selected_dqis, if not I can remove it

ashmortar: actually, once all events are guaranteed to carry selected_data_quality_issues, this branch becomes dead code. But it's load-bearing now for backward compatibility with old events.

alena-hutchinson: Yeah, exactly. I'll update the FF cleanup ticket with a link to this thread so whoever takes it knows to remove this branch since it wouldn't come up in a typical FF search.

alena-hutchinson (approved):

PR #1171, opened 2026-04-20
Add experimental generate-expectations event and action shell

joshua-stauffer:
The ExpectAI v2 rollout needs a dedicated event type so the new pipeline can be exercised end-to-end alongside the existing v1 flow without disturbing it. This change lands the event schema and an empty action shell so the routing is in place; the actual pipeline wiring comes in a later ticket.

Specifically, this adds `GenerateExpectationsExperimentalEvent` in `agent/models.py` (discriminator `generate_expectations_experimental_action.received`, with `datasource_name`, `data_asset_name`, `batch_definition_name`, and optional `batch_parameters`) and a corresponding `GenerateExpectationsExperimentalAction` registered for GX Core major version 1. The action is intentionally a no-op that returns a successful `ActionResult` with no created resources. It is imported from `actions/__init__.py` so registration happens at package import time, mirroring the existing pattern.

Unit tests cover the event being discovered by the dynamic Event union, parsing via the discriminated union, the action returning an empty `ActionResult`, and the event handler dispatching the event to the new action. Lint, mypy, and the full agent test suite pass locally (377 passed, 1 skipped).


ashmortar (approved):

ashmortar:
@copilot resolve the merge conflicts in this pull request

